PR #2, opened 2019-04-29
Ingest users and groups

Andrew8xx8:


package.json
```diff
@@ -17,18 +17,26 @@
...
     "stop:graph": "docker stop jupiterone-local-graph",
-    "prebuild": "yarn test",
-    "build": "rm -rf dist/* && yarn build:types && yarn build:src && yarn build:docs",
+    "prebuild": "rm -rf dist/*",
+    "build": "yarn build:types && yarn build:src && yarn build:files",
     "build:src": "babel src --extensions '.ts' --out-dir dist",
-    "build:docs": "ts-node tools/docs.ts",
+    "build:files": "ts-node tools/copy-dist-files.ts",
     "build:types": "tsc --project tsconfig.dist.json --declaration --emitDeclarationOnly",
+    "build:publish": "yarn build && cd dist && yarn publish --access public",
```
aiwilliams: These `prebuild`, `build`, `build:files`, and `build:publish` lines need to be reverted, as the previous forms are what make the Travis build and NPM deploy from Travis work.

package.json
```diff
@@ -17,18 +17,26 @@
...
     "lint": "yarn lint:ts && yarn lint:md",
     "lint:md": "remark .",
     "lint:ts": "tslint --project .",
     "format": "yarn prettier --write '**/*.{ts,js,json,md}'",
     "test": "jest"
   },
-  "dependencies": {},
+  "dependencies": {
+    "@microsoft/microsoft-graph-client": "^1.5.2",
+    "bunyan": "^1.8.12",
```
aiwilliams: This is in `devDependencies` already, and is only used in dev, so should be removed here.

tools/copy-dist-files.ts
```diff
@@ -0,0 +1,24 @@
+import fs from "fs";
```
aiwilliams: This file should be deleted, as it will no longer be used once `package.json` comments are processed to revert a few scripts.

src/converters/entities/UserEntityConverter.ts
```diff
@@ -0,0 +1,30 @@
...
+
+import { generateEntityKey } from "../../utils/generateKeys";
+
+export function createUserEntities(data: User[]): UserEntity[] {
+  return data.map(u => {
+    const user: UserEntity = {
+      _class: USER_ENTITY_CLASS,
+      _key: generateEntityKey(USER_ENTITY_TYPE, u.id),
+      _type: USER_ENTITY_TYPE,
+      displayName: u.displayName || "",
```
aiwilliams: I hope I am not being inconsistent, but I think we want properties to be `undefined` on the entities if they are not defined on the provider data, or there is not a non-blank value we can derive from other properties.

src/converters/relationships/GroupMemberRelationshipConverter.ts
```diff
@@ -0,0 +1,76 @@
...
+    [] as GroupUserRelationship[],
+  );
+}
+
+function createRelationship(group: Group, member: GroupMember) {
+  const childKey = getMemberKey(member);
+  const parentKey = generateEntityKey(GROUP_ENTITY_TYPE, group.id);
+
+  const relationshipType =
+    member["@odata.type"] === MemberType.USER
```
aiwilliams: I wonder if it would be safer to specifically check that type is `USER` or type is `GROUP` else log something and skip this entity 🤔 Is there any indication in the docs that this could have other values some day? Anyway, just a thought, not worth holding up this PR.

PR #17, opened 2019-05-22
Feature/error handling

Andrew8xx8:


aiwilliams:
> The message of the error should be a summary from the response

Here is an example of building an error when `fetch` fails: https://github.com/JupiterOne/jupiter-integration-jamf/blob/master/src/jamf/JamfClient.ts#L64 Note that it constructs as meaningful a message as possible for errors it knows about, and uses `response.statusText` for unknown errors.

src/azure/AzureClient.ts
```diff
@@ -57,27 +69,57 @@ export default class AzureClient {
...
+      const { value } = await this.makeRequest(
+        `${this.host}/groups/`,
+        Method.GET,
+      );
+
+      return value;
+    } catch (error) {
+      if (error instanceof AzureClientError) {
+        this.logger.info(error.message);
+        return [];
```
aiwilliams: When an API error occurs, and we want to continue processing other entities, then we should not re-throw of course, but we have to decide what to do in the synchronization to avoid deleting existing data in the graph. The pattern we've adopted to date is to return `undefined`, and the caller should interpret that to mean, do not attempt to synchronize this kind of data.

src/azure/AzureClient.ts
```diff
@@ -57,27 +69,57 @@ export default class AzureClient {
...
+      const { value } = await this.makeRequest(
+        `${this.host}/groups/${groupId}/members`,
+        Method.GET,
+      );
+
+      return value;
+    } catch (error) {
+      if (error instanceof AzureClientError) {
+        this.logger.info(error.message);
+        return [];
```
aiwilliams: Same comments here:

```
this.logger.warn({err: error}, "azure.fetchMembers failed");
```

And:

```
return undefined;
```

src/azure/AzureClientUnexpectedError.test.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    return azure;
+  }
+
+  test("fetchUsers with a error", async () => {
+    const client = await getClient();
+    try {
+      await client.fetchUsers();
+    } catch (e) {
+      expect(e instanceof Error).toBe(true);
```
aiwilliams: I'd recommend checking the expected type of the error.
```suggestion
      expect(e).toInstanceOf(IntegrationError);
```
Or you may like (I think this should work):
```suggestion
      await expect(client.fetchUsers()).rejects.toBeInstanceOf(IntegrationError);
```

PR #18, opened 2019-05-28
Normalize user data and ingest manager info

Andrew8xx8:
**IMPORTANT note** Merge #17 first


src/azure/fetchAzureData.ts
```diff
@@ -9,11 +15,15 @@ export default async function fetchAzureData(
     client.fetchUsers(),
   ]);
 
+  const usersManagers = await Promise.all(
```
aiwilliams: Does the users' manager not appear in the regular set of users? Is that why the integration must build User entities for managers, instead of counting on regular user processing getting them into the graph? Perhaps there are cases where we need both, in which case, it would seem we need a test case to reflect that.

Andrew8xx8: This is not creating additional User records. It just attaches Manager information to a matched user record.

src/azure/AzureClientUnexpectedError.test.ts
```diff
@@ -0,0 +1,77 @@
...
+  });
+
+  test("fetchUserManager with a error", async () => {
+    const client = await getClient();
+    try {
+      await client.fetchUserManager({
+        id: "89fac263-2430-48fd-9278-dacfdfc89792",
+      });
+    } catch (e) {
+      expect(e instanceof Error).toBe(true);
```
aiwilliams: Is this what is to be expected, or should it be an `undefined` value?

aiwilliams:
This has grown too stale. Will track review task in project planning.

PR #23, opened 2019-09-12
Refactor to steps

aiwilliams:
A major refactoring to spread work across a number of steps for using in a Step Functions state machine. This will spread the work out and allow for iterating data across a number of Lambda invocations.

Changes to produced graph:

* There were two relationships created between a group and user, now there is only one, `Group -HAS-> User`
* There is now only a single group -> member relationship type
* Group members that are not a User or Group are now ingested (leveraging mapped relationships to make this happen)
* Lists are paginated so that complete sets are ingested


src/azure/AzureClient.ts
```diff
@@ -64,65 +91,72 @@ export default class AzureClient {
...
+        );
       }
-      this.logger.warn({ err: error }, "azure.fetchMembers failed");
-      return undefined;
     }
   }
 
+  /**
+   * Pagination: https://docs.microsoft.com/en-us/graph/paging
+   * Throttling with retry after: https://docs.microsoft.com/en-us/graph/throttling
```
aiwilliams: We need to handle throttling still!

src/azure/AzureClient.ts
```diff
@@ -64,65 +91,72 @@ export default class AzureClient {
...
+   * Batching requests: https://docs.microsoft.com/en-us/graph/json-batching
```
aiwilliams: Batching would be a pretty useful feature when we get around to asking for object details, don't see much value in these listing calls.

fomentia:
dab

PR #24, opened 2019-09-13
Ingest compute resources

aiwilliams:


src/jupiterone/entities.ts
```diff
@@ -59,3 +63,13 @@ export interface GroupMemberEntity extends EntityFromIntegration {
...
 }
+
+export interface VirtualMachineEntity extends EntityFromIntegration {
+  vmSize: string | undefined;
+  type: string | undefined;
+  vmId: string | undefined;
+  location: string;
+}
+
+export const VIRTUAL_MACHINE_ENTITY_TYPE = "azurerm_virtual_machine";
```
erkangz: I know we try to stay consistent with TF types but I think their azure naming is too complicated that would make queries in J1 unnecessarily difficult.  I think `azure_vm` would be much better here

aiwilliams: 😅Okay, great! I like `azure_vm` much better.

aiwilliams:
I think this is ready to go. I've made it so that it does not require the `subscriptionId` - it will not ingest compute resources if it is not in the config - since we have Azure configurations already which do not have the value.

jest.config.js
```diff
@@ -22,6 +22,18 @@ module.exports = {
...
+      branches: 100,
+      functions: 100,
+      lines: 100,
+    },
+    "src/azure/resource-manager": {
+      statements: 100,
+      branches: 100,
+      functions: 100,
+      lines: 100,
+    },
```
fomentia: I wonder if these coverage statements should be inverted, I.E. exclude things you don't care to cover instead of including everything you do want to cover. It seems like exclusion should be more exceptional than inclusion.

aiwilliams: My goal was to ensure that the things at 100% remain that way, and at the start, that meant a smaller list when I could point to complete directories. I believe a tipping point is on the horizon so that I can do as you suggest 👍

test/helpers/polly.ts
```diff
@@ -0,0 +1,115 @@
...
+import { Har } from "har-format";
+import { gunzipSync } from "zlib";
+
+import NodeHttpAdapter from "@pollyjs/adapter-node-http";
+import { Polly, PollyConfig } from "@pollyjs/core";
+import FSPersister from "@pollyjs/persister-fs";
+
+Polly.register(NodeHttpAdapter);
+
+class SafeHeadersFSPersister extends FSPersister {
```
fomentia: This class name is a bit misleading now but nbd tbh

test/helpers/pollyjs__persister-fs.d.ts
```diff
@@ -0,0 +1,9 @@
+declare module "@pollyjs/persister-fs" {
+  import { Polly } from "@pollyjs/core";
+  import Persister from "@pollyjs/persister";
+
+  export default class FSPersister extends Persister {
+    constructor(polly: Polly);
+    public saveRecording(recordingId: number, data: any): void;
```
fomentia: `data` here should be `Har`

package.json
```diff
@@ -29,9 +29,12 @@
...
   },
   "dependencies": {
+    "@azure/arm-compute": "^11.0.0",
+    "@azure/ms-rest-nodeauth": "^2.0.5",
     "@jupiterone/jupiter-managed-integration-sdk": "^24.2.0",
-    "@microsoft/microsoft-graph-client": "^1.5.2",
-    "node-fetch": "^2.3.0"
+    "@microsoft/microsoft-graph-client": "^2.0.0",
+    "cross-fetch": "^3.0.4",
+    "lodash.map": "^4.6.0"
```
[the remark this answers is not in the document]
aiwilliams: Yah, I started writing a function that I could test for the pattern `collection ? collection.map(...) : undefined`, so that the branch would not existing in the converter code, which would require a number of lines of test code to satisfy the coverage tool on that `undefined` collection branch. I decided to swallow the 💊

src/azure/resource-manager/iterateVirtualMachines.ts
```diff
@@ -0,0 +1,21 @@
...
+import { VirtualMachine } from "@azure/arm-compute/esm/models";
+
+export default async function iterateVirtualMachines(
+  client: ComputeManagementClient,
+  callback: (vm: VirtualMachine) => void,
+) {
+  let nextLink: string | undefined;
+  do {
+    const response = nextLink
+      ? /* istanbul ignore next: how to control page number? */
```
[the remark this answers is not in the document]
aiwilliams: The coverage tool complained that I was not testing pagination. I don't see a way in the client library to control the page number of objects, so that I could write a test with 2 VMs, 1 per page, to exercise this branch 😕

PR #26, opened 2019-09-20
Ingest more network resources

aiwilliams:
This is a work in progress, but wanted to demonstrate progress :) The `@azure/*` packages are many. Note here how the compute resources are in one package while network resources are accessed through another. These details are getting wrapped up in a `ResourceManagerClient` so that we don't have to know about all that in the synchronizers.


src/jupiterone/entities.ts
```diff
@@ -14,32 +14,32 @@ export const GROUP_ENTITY_CLASS = "UserGroup";
...
-  classification?: string;
-  createdDateTime?: number;
-  description?: string;
-  displayName?: string;
-  mail?: string;
-  mailEnabled?: boolean;
-  mailNickname?: string;
-  renewedDateTime?: number;
-  securityEnabled?: boolean;
+  deletedDateTime: number | undefined;
```
fomentia: What is the difference here? Does `number | undefined` force you to explicitly declare `deletedDateTime: undefined` when you create one of these entities?

aiwilliams: Correct! It's a trick I picked up from @erkangz, and I think a good idea for these converters to be forced to transmit something into the structure.

erkangz: I think we should add `createdOn` and `deletedOn` as normalized properties for querying

src/converters/resources.ts
```diff
@@ -1,26 +1,93 @@
...
+    _key: data.id as string,
+    _type: PUBLIC_IP_ADDRESS_ENTITY_TYPE,
+    _class: PUBLIC_IP_ADDRESS_ENTITY_CLASS,
+    _rawData: [{ name: "default", rawData: data }],
+    resourceGuid: data.resourceGuid,
+    displayName: data.name,
+    type: data.type,
+    region: data.location,
+    publicIp: data.ipAddress,
+    publicIpAddress: data.ipAddress,
```
erkangz: let's add `public: true` here

PR #29, opened 2019-09-27
Conditional execution of AD ingestion steps

aiwilliams:


aiwilliams:
Looking at https://docs.aws.amazon.com/step-functions/latest/dg/amazon-states-language-choice-state.html, this cannot be an Array of `disabledSteps`, as there are not Choice state comparison operators that support that. Will instead need to be able to look up a step's runtime configuration, such as:

```json
{
  "Variable": "$.step.enabled",
  "NumericEquals": 1,
  "Next": "Step Name Init"
}
```

aiwilliams:
Note that local execution does not regard step disablement at this time.

src/index.ts
```diff
@@ -40,12 +53,52 @@ export const stepFunctionsInvocationConfig: IntegrationInvocationConfig = {
...
+  getStepStartStates: (
+    executionContext: IntegrationExecutionContext,
+  ): IntegrationStepStartStates => {
+    const config = (executionContext.instance.config ||
+      {}) as AzureIntegrationInstanceConfig;
+
+    const activeDirectory = { disabled: !config.ingestActiveDirectory };
+    const resourceManager = {
+      disabled: typeof config.subscriptionId !== "string",
+    };
```
fomentia: The names of these are a bit confusing to me... Maybe they could be named something like `activeDirectoryDisabled`?

src/index.ts
```diff
@@ -40,12 +53,52 @@ export const stepFunctionsInvocationConfig: IntegrationInvocationConfig = {
...
+
+    const states: IntegrationStepStartStates = {};
+
+    if (activeDirectory.disabled) {
+      Object.assign(states, {
```
fomentia: Is there any reason to use `Object.assign` over the spread operator? I typically prefer the spread operator as it forces me to make the variable I'm going to mutate a `let` variable instead of a `const` (to indicate that it actually will be mutated).

src/index.ts
```diff
@@ -117,27 +175,22 @@ export const stepFunctionsInvocationConfig: IntegrationInvocationConfig = {
...
-            ) {
-              return synchronizeComputeResources(
-                initializeContext(executionContext),
-              );
-            } else {
-              return Promise.resolve({});
-            }
+            return synchronizeComputeResources(
+              initializeContext(executionContext),
+            );
```
fomentia: Wow! So much better! :grin:

fomentia:
All feedback optional / just questions, LGTM!

src/index.ts
aiwilliams: I did that name at first, but it was confusing to me because the object is state for `activeDirectory` steps, which in my mind could at some point include properties in addition to `disabled` 🤷‍♂

src/index.ts
aiwilliams: That's a great reason to use `let` and reassign with spread operator. I've made that change. Thanks!

PR #38, opened 2019-11-07
Ingest vnet, subnet, security groups

aiwilliams:


src/converters/resources.ts
```diff
@@ -103,6 +116,82 @@ export function createVirtualMachineEntity(
...
+        webLink: webLinker.portalResourceUrl(data.id),
+        CIDR,
+        public: false,
+        internal: true,
+        region: data.location,
+        resourceGroup: resourceGroupName(data.id),
+      },
+      tagProperties: ["environment"],
+    },
+  });
```
aiwilliams: Note that a TS type is not being maintained for these generated entities. Adding these properties was driven by the converter tests and schema validation that occurs in `createIntegrationEntity`. I do expect to generate TS types from the schemas, but so far I haven't really missed having the types (which we typically hand-code).

src/converters/resources.test.ts
```diff
@@ -257,6 +265,333 @@ describe("createVirtualMachineEntity", () => {
...
+      _rawData: [{ name: "default", rawData: data }],
+      name: "j1dev",
+      displayName: "j1dev",
+      resourceGroup: "j1dev",
+      region: "eastus",
+      environment: "j1dev",
+      webLink: webLinker.portalResourceUrl(
+        "/subscriptions/dccea45f-7035-4a17-8731-1fd46aaa74a0/resourceGroups/j1dev/providers/Microsoft.Network/networkSecurityGroups/j1dev",
+      ),
+      category: "network",
```
aiwilliams: I guess this means that `Firewall.category` should be an Array, with ["network", "host"] when a SG is used on a subnet AND a NIC?

src/converters/resources.ts
```diff
@@ -103,6 +116,82 @@ export function createVirtualMachineEntity(
...
+): EntityFromIntegration {
+  const CIDR =
+    data.addressSpace &&
+    data.addressSpace.addressPrefixes &&
+    data.addressSpace.addressPrefixes.length > 0 &&
+    data.addressSpace.addressPrefixes[0];
+
+  return createIntegrationEntity({
+    entityData: {
+      source: data,
```
aiwilliams: Note that `source` becomes `entity._rawData = [{ name: "default", rawData: source }]`. It is possible to add additional raw data through the `assign` argument. `createIntegrationEntity` will validate `_rawData` to ensure there are no duplicate entries.

src/converters/resources.ts
```diff
@@ -103,6 +116,82 @@ export function createVirtualMachineEntity(
...
+
+export function createVirtualNetworkEntity(
+  webLinker: AzureWebLinker,
+  data: VirtualNetwork,
+): EntityFromIntegration {
+  const CIDR =
+    data.addressSpace &&
+    data.addressSpace.addressPrefixes &&
+    data.addressSpace.addressPrefixes.length > 0 &&
+    data.addressSpace.addressPrefixes[0];
```
aiwilliams: We need to move to latest TS to get `data.addressSpace?.addressPrefixes?[0]` 🤤

src/converters/resources.ts
```diff
@@ -103,6 +116,82 @@ export function createVirtualMachineEntity(
...
+}
+
+export function createVirtualNetworkEntity(
+  webLinker: AzureWebLinker,
+  data: VirtualNetwork,
+): EntityFromIntegration {
+  const CIDR =
+    data.addressSpace &&
+    data.addressSpace.addressPrefixes &&
+    data.addressSpace.addressPrefixes.length > 0 &&
```
aiwilliams: Notice that this is an Array. A network can have multiple CIDRs. @erkangz, I think we'll need to change the data model to support an Array. Sound good?

src/converters/resources.ts
```diff
@@ -103,6 +116,82 @@ export function createVirtualMachineEntity(
...
+        displayName: CIDR ? `${data.name} (${CIDR})` : data.name,
+        webLink: webLinker.portalResourceUrl(data.id),
+        CIDR,
+        public: false,
+        internal: true,
+        region: vnet.location,
+        resourceGroup: resourceGroupName(data.id),
+        environment: vnet.tags && vnet.tags["environment"],
+      },
+      tagProperties: ["environment"],
```
aiwilliams: This should be removed. The `Subnet` included in the `VirtualNetwork.subnets` does not include any tags, so it is transferred from the vnet above.

src/converters/resources.test.ts
erkangz: Sounds good

src/azure/utils.test.ts
```diff
@@ -0,0 +1,34 @@
...
+    expect(
+      resourceGroupName(
+        "/subscriptions/dccea45f-7035-4a17-8731-1fd46aaa74a0/resourceGroups/J1DEV/providers/Microsoft.Network/networkInterfaces/j1dev",
+      ),
+    ).toEqual("j1dev");
+    expect(
+      resourceGroupName(
+        "/subscriptions/dccea45f-7035-4a17-8731-1fd46aaa74a0/resourceGroups/🤔/providers/Microsoft.Network/networkInterfaces/j1dev",
+      ),
+    ).toEqual("🤔");
```
fomentia: :thinking:

src/azure/utils.ts
```diff
@@ -12,11 +12,24 @@ const resourceGroupRegex = /\/resourceGroups\/(.*?)\//;
  * @param id the resource id, which is expected to include the resource group
  * @returns the resource group name lowercased, or `undefined` when not found
  */
-export function resourceGroup(id: string | undefined): string | undefined {
+export function resourceGroupName(
+  id: string | undefined,
+  required = false,
```
fomentia: I think we'd want to add this property to this list of `@param`s above

src/converters/resources.ts
fomentia: @aiwilliams Wait, does the latest TS have that feature? I thought it was just in proposal status. :flushed:

src/converters/resources.ts
fomentia: Yeah, the types aren't quite as necessary when you have schema validation and you also don't have to type out every single property. :heart:

fomentia:
The new helper function really makes the converters a lot cleaner. Me gusta! :100:

aiwilliams:
Screen Shot 2019-11-08 at 4 41 46 PM

PR #39, opened 2019-11-21
Ingest storage accounts

aiwilliams:
Ingests Blob containers. Depends on a release of the integration SDK. Still need to handle File, Queue, and Table storage services, though that should be getting easier to do with these changes setting up for it. Also need to get more details about the Blob service containers (like S3 buckets), such as policies etc.

Screen Shot 2019-11-22 at 2 31 07 PM

### DIFF
--- a/docs/tagging.md
+++ b/docs/tagging.md
@@ -1,6 +1,6 @@
 # Microsoft Resource Manager Tag Support
 
-Grouping resouces is important for accomplishing a number of infrastructure
+Grouping resources is important for accomplishing a number of infrastructure
 management tasks. This is supported in Azure using resource groups and tagging.
 The Azure [Tags How to][1] states:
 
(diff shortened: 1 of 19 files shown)


docs/tagging.md
```diff
@@ -0,0 +1,21 @@
+# Microsoft Resource Manager Tag Support
+
+Grouping resouces is important for accomplishing a number of infrastructure
+management tasks. This is supported in Azure using resource groups and tagging.
+The Azure [Tags How to][1] states:
+
+> Tags enable you to retrieve related resources from different resource groups.
+
+See [Tag Support][2] for details on which resources support tagging, because not
+all of them do 😳. This means that it will not be possible to leverage tags as
```
fomentia:
```suggestion
all of them do. 😳 This means that it will not be possible to leverage tags as
```
No emojis before the period. :triumph:

docs/tagging.md
```diff
@@ -0,0 +1,21 @@
+# Microsoft Resource Manager Tag Support
+
+Grouping resouces is important for accomplishing a number of infrastructure
```
fomentia:
```suggestion
Grouping resources is important for accomplishing a number of infrastructure
```

PR #50, opened 2020-01-27
Additional Logs

fomentia:
This should address all of the logs I said I would add in the related Jira ticket.

### DIFF
--- a/src/azure/resource-manager/BunyanLogPolicy.ts
+++ b/src/azure/resource-manager/BunyanLogPolicy.ts
@@ -42,7 +42,8 @@ export class BunyanLogPolicy extends BaseRequestPolicy {
       const logData = {
         url: httpRequest.url,
         status: response.status,
-        headers: response.parsedHeaders,
+        requestHeaders: response.request.headers.toString(),
+        responseHeaders: response.headers.toString(),
       };
 
       this.logger.info(logData, "Received response from Azure API");
--- a/src/azure/resource-manager/BunyanLogPolicy.ts
+++ b/src/azure/resource-manager/BunyanLogPolicy.ts
@@ -42,7 +42,7 @@ export class BunyanLogPolicy extends BaseRequestPolicy {
       const logData = {
         url: httpRequest.url,
         status: response.status,
-        responseHeaders: response.headers.toString(),
+        responseHeaders: response.headers.rawHeaders(),
       };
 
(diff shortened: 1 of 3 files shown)


src/azure/resource-manager/BunyanLogPolicy.ts
```diff
@@ -42,7 +42,8 @@ export class BunyanLogPolicy extends BaseRequestPolicy {
       const logData = {
         url: httpRequest.url,
         status: response.status,
-        headers: response.parsedHeaders,
+        requestHeaders: response.request.headers.toString(),
```
aiwilliams: I'm concerned about this one because I expect it includes sensitive values. The first request to obtain an OAuth token would include the API secret, right?

src/synchronizers/synchronizeStorageAccounts.ts
```diff
@@ -164,13 +171,20 @@ async function synchronizeStorageAccount(
       const synchronizer = storageServiceSynchronizers[s];
       if (synchronizer) {
         logger.info(logInfo, "Processing storage account service...");
-        operationsResults.push(
-          await synchronizer(
-            executionContext,
-            accountEntity,
-            webLinker,
+        const operations = await synchronizer(
```
aiwilliams: I think it would be good to name this something other than `operations` because I thought this was logging all the operations. How about `operationsSummary`, the name used later in the `logger.info()` call?

src/azure/resource-manager/BunyanLogPolicy.ts
fomentia: I added this to help debug: https://lifeomic.atlassian.net/browse/JPT-2969

I'll remove it for now though.

aiwilliams: Oh, interesting. Perhaps we could start by logging the header keys so we know what's included, and then log values other than sensitive ones such as `Authorization`. I know the feedback cycles are way too long to love that approach, but we have to keep secrets out of the logs.

fomentia: Drat! I wish I had seen this comment before I merged, that's a great idea. Darn GitHub resolved conversations! :fist_raised:

fomentia: I'll create another PR to this project real quick with that change so we don't have to create two upgrade PRs in the deployment project.

PR #52, opened 2020-02-03
Avoid fetching whole graph more than once

fomentia:


docker-compose.yml
```diff
@@ -18,6 +18,7 @@ services:
       - graph
   graph:
     image: tinkerpop/gremlin-server:3.4
+    user: "1000"
```
fomentia: I know we've had a discussion with @erichs about this but I can't remember what we decided. I'll try to find it...

src/synchronizers/synchronizeStorageAccounts.ts
```diff
@@ -39,12 +98,38 @@ export default async function synchronizeStorageAccounts(
...
+          "HAS",
+          STORAGE_BLOB_SERVICE_ENTITY_TYPE,
+          STORAGE_CONTAINER_ENTITY_TYPE,
+        ),
+      ),
+    ]),
+  );
+
+  executionContext.logger.info(
+    {},
```
aiwilliams: Would be pretty great to include counts in this object.

docker-compose.yml
fomentia: https://bitbucket.org/lifeomic/jupiter-integration-aws/pull-requests/479/support-cross-platform-docker-env

fomentia:
OK, it's doing everything it needs to do, but I think the code can be made more readable by breaking it apart some (functions are way too long). Maybe should separate it into multiple files as well.

fomentia:
Ugh. Feel free to review now, I may not update it. I hate JavaScript.

fomentia:
@aiwilliams Pls review. Are there places where I should add more logs?

src/synchronizers/synchronizeStorageAccounts/index.ts
```diff
@@ -0,0 +1,298 @@
...
+      "Finished getting old container entity and reassigning old data",
+    );
+
+    const [
+      serviceContainerRelationship,
+      oldServiceContainerRelationship,
+      serviceContainerOldData,
+    ] = entityOrRelationship({
+      newEntityConstructor: () => {
+        return createIntegrationRelationship({
```
aiwilliams: I think these can be created and assigned, then passed in without having to just return and destructure it. The function is not deferred or otherwise used in interesting ways.

src/synchronizers/synchronizeStorageAccounts/index.ts
```diff
@@ -0,0 +1,298 @@
...
+  const newContainerEntities: EntityFromIntegration[] = [];
+  const oldServiceContainerRelationships: IntegrationRelationship[] = [];
+  const newServiceContainerRelationships: IntegrationRelationship[] = [];
+
+  await azrm.iterateStorageBlobContainers(storageAccount, e => {
+    const [
+      containerEntity,
+      oldContainerEntity,
+      containerOldData,
+    ] = entityOrRelationship({
```
aiwilliams: This name does't capture all that is happening, particularly the fact that `oldData` is being worked off.

src/synchronizers/synchronizeStorageAccounts/index.ts
```diff
@@ -0,0 +1,298 @@
...
+      newEntityConstructor: () => {
+        return createStorageContainerEntity(webLinker, storageAccount, e);
+      },
+      oldData,
+      oldDataGetter: getOldContainerEntity,
+    });
+    oldData = containerOldData;
```
aiwilliams: This is confusing. Perhaps use `let oldData: OldData` before this assignment, so that it can be reassigned in the destructure assignment?

src/synchronizers/synchronizeStorageAccounts/index.ts
```diff
@@ -0,0 +1,265 @@
...
+      to: containerEntity,
+    });
+    let oldServiceContainerRelationship;
+    [
+      oldServiceContainerRelationship,
+      oldData,
+    ] = popOldServiceContainerRelationship(
+      oldData,
+      serviceContainerRelationship,
+    );
```
aiwilliams: You know what would make this just a little bit better?
```
    let oldServiceContainerRelationship;
    [
      oldData,
      oldServiceContainerRelationship,
    ] = popOldServiceContainerRelationship(
      oldData,
      serviceContainerRelationship,
    );
```

fomentia: @aiwilliams I considered that, but then thought it would be weird for pop to _not_ return the popped item as the first item in the tuple. Although, now that I'm thinking about it again, it makes sense and is more readable, so I'll update.

PR #53, opened 2020-02-04
Add support for SQL servers and databases

erkangz:


src/converters/relationships.ts
```diff
@@ -254,3 +257,19 @@ export function createVirtualMachineNetworkInterfaceRelationship(
...
+  database: EntityFromIntegration,
+): IntegrationRelationship {
+  return createIntegrationRelationship({
+    _class: SQL_SERVER_DATABASE_RELATIONSHIP_CLASS,
+    fromKey: server._key,
+    fromType: server._type,
+    toKey: database._key,
+    toType: database._type,
+    properties: {
+      _type: SQL_SERVER_DATABASE_RELATIONSHIP_TYPE,
```
aiwilliams: Since this is a new relationship, it would be great to avoid passing this and allow the `createIntegrationRelationship()` function to build the `_type` using the `fromType`, `_class`, `toType` values, seen here: https://bitbucket.org/lifeomic/jupiter-managed-integration-sdk/src/70e47b55361ac2884edfd1238e6dc8c2b075e5e1/src/data-model/createIntegrationRelationship.test.ts#lines-75

The only reason the other functions are providing it is because I adopted `createIntegrationRelationship` in this integration and did not want to change the `_type` values.

src/jupiterone/relationships.ts
```diff
@@ -29,3 +29,7 @@ export const VIRTUAL_MACHINE_NIC_RELATIONSHIP_CLASS = "USES";
 export const VIRTUAL_MACHINE_PUBLIC_IP_ADDRESS_RELATIONSHIP_TYPE =
   "azure_vm_uses_public_ip";
 export const VIRTUAL_MACHINE_PUBLIC_IP_ADDRESS_RELATIONSHIP_CLASS = "USES";
+
+export const SQL_SERVER_DATABASE_RELATIONSHIP_TYPE =
+  "azure_sql_server_has_database";
```
aiwilliams: This can be assigned using `generateRelationshipType()` from https://bitbucket.org/lifeomic/jupiter-managed-integration-sdk/src/70e47b55361ac2884edfd1238e6dc8c2b075e5e1/src/data-model/util.ts#lines-8. For example, https://github.com/JupiterOne/graph-crowdstrike/blob/master/src/jupiterone/converters.ts#L118. I don't love that we include the class in the `_type`, because it means that when we change the relationship class the type is different, but we keep doing that, so that is what this generator does.

Note that once we implement the new approach to synchronization, where we iterate the data from J1, we will not need to generate relationship type constants for looking up stuff like we have to do here.

src/converters/relationships.ts
aiwilliams: To clarify, I meant that I did not want to have to write up code to "deprecate" (delete) relationships having the existing type value.

src/jupiterone/relationships.ts
erkangz: the generated relationship type is something like `azure_sql_server_has_azure_sql_database` which is super wordy.  maybe that's ok?  If we later allow querying for edges directly, this may matter

erkangz: ok, made the SDK smarter to generate type strings without duplicating the provider and service portion (e.g. `azure_sql_`) in the latest version

src/jupiterone/relationships.ts
```diff
@@ -1,31 +1,22 @@
...
-  "azure_vm_uses_network_interface";
-export const VIRTUAL_MACHINE_NIC_RELATIONSHIP_CLASS = "USES";
+export const VIRTUAL_MACHINE_NIC_RELATIONSHIP_TYPE = "azure_vm_uses_nic";
 
 export const VIRTUAL_MACHINE_PUBLIC_IP_ADDRESS_RELATIONSHIP_TYPE =
   "azure_vm_uses_public_ip";
-export const VIRTUAL_MACHINE_PUBLIC_IP_ADDRESS_RELATIONSHIP_CLASS = "USES";
+
+export const SQL_SERVER_DATABASE_RELATIONSHIP_TYPE =
+  "azure_sql_server_has_database";
```
aiwilliams: What do you think about making this:
```
export const SQL_SERVER_DATABASE_RELATIONSHIP_TYPE = generateRelationshipType(DataModel.RelationshipClass.HAS, SQL_SERVER_ENTITY_TYPE, SQL_DATABASE_ENTITY_TYPE)
```
That will ensure that it generates whatever is being produced in `createIntegrationRelationship()`.

src/synchronizers/syncronizeComputeResources.ts
```diff
@@ -127,6 +127,16 @@ export default async function synchronizeComputeResources(
...
+  // This has been changed to "azure_vm_uses_nic"
+  const DEPRECATED_VIRTUAL_MACHINE_NIC_RELATIONSHIP_TYPE =
+    "azure_vm_uses_network_interface";
+
+  oldVMNicRelationships.push(
+    ...(await graph.findRelationshipsByType(
+      DEPRECATED_VIRTUAL_MACHINE_NIC_RELATIONSHIP_TYPE,
+    )),
+  );
+
```
aiwilliams: Excellent, thanks for getting this in here. I didn't notice any other deprecated types 👍

PR #55, opened 2020-02-05
Add support for MariaDB and PostgreSQL

erkangz:


src/azure/resource-manager/ResourceManagerClient.ts
```diff
@@ -208,6 +220,70 @@ export default class ResourceManagerClient {
...
 
+  public async iterateMariaDbServers(
+    callback: (s: MariaDBServer) => void,
+  ): Promise<void> {
+    const client = ((await this.getAuthenticatedClient(
+      MariaDBManagementClient as any,
+    )) as unknown) as MariaDBManagementClient;
+    const servers = await client.servers.list();
+    for (const server of servers) {
+      callback(server);
```
aiwilliams:
```suggestion
      await callback(server);
```

src/azure/resource-manager/ResourceManagerClient.ts
```diff
@@ -208,6 +220,70 @@ export default class ResourceManagerClient {
...
+
+  public async iteratePostgreSqlServers(
+    callback: (s: PostgreSQLServer) => void,
+  ): Promise<void> {
+    const client = ((await this.getAuthenticatedClient(
+      PostgreSQLManagementClient as any,
+    )) as unknown) as PostgreSQLManagementClient;
+    const servers = await client.servers.list();
+    for (const server of servers) {
+      callback(server);
```
aiwilliams:
```suggestion
      await callback(server);
```

src/converters/resources/database.ts
```diff
@@ -56,6 +56,8 @@ export function createDbServerEntity(
       source: data,
       assign: {
         ...convertProperties(data),
+        ...convertProperties((data as any).sku, { prefix: "sku" }),
+        ...convertProperties((data as any).storageProfile),
```
aiwilliams: Does `convertProperties(data)` avoid flattening object properties, so that there will not be a `sku: JSON.stringify(data.sku)` or `storageProfile: JSON.stringify(data.storageProfile)`?

src/synchronizers/syncronizeDatabaseResources.ts
```diff
@@ -44,27 +44,17 @@ export default async function synchronizeDatabaseResources(
...
-  );
-
   return {
     operations: summarizePersisterOperationsResults(
-      operationsResultMySQL,
-      operationsResultSQL,
+      await synchronize(executionContext, webLinker, DatabaseType.MariaDB),
+      await synchronize(executionContext, webLinker, DatabaseType.MySQL),
+      await synchronize(executionContext, webLinker, DatabaseType.PostgreSQL),
+      await synchronize(executionContext, webLinker, DatabaseType.SQL),
```
aiwilliams: This is good I think, to avoid a `Promise.all()`, so that we don't hammer the storage APIs 👍

src/converters/resources/database.ts
erkangz: only if you set the option `...convertProperties(data, { stringifyObject: true }),`. by default it skips them

src/synchronizers/syncronizeDatabaseResources.ts
erkangz: yep

PR #58, opened 2020-02-17
Do not page when access to group members is denied

fomentia:
User configuration error, should be reported in the job log but not page.

### DIFF
--- a/src/azure/AzureClient.test.ts
+++ b/src/azure/AzureClient.test.ts
@@ -12,11 +12,11 @@ const DIRECTORY_ID =
 
 const logger = createTestLogger();
 
-function createAzureClient() {
+function createAzureClient(): AzureClient {
   return new AzureClient(CLIENT_ID, CLIENT_SECRET, DIRECTORY_ID, logger);
 }
 
-function createClientWithAuth() {
+function createClientWithAuth(): AzureClient {
   nock("https://login.microsoftonline.com")
     .post(`/${DIRECTORY_ID}/oauth2/v2.0/token`)
     .reply(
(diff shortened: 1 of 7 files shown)


src/synchronizers/synchronizeGroupMembers.ts
```diff
@@ -22,8 +23,11 @@ export default async function synchronizeGroupMembers(
 
   const groupsState = await groupsCache.getState();
   if (!groupsState || !groupsState.groupMembersFetchCompleted) {
-    throw new IntegrationError(
-      "Group members fetching did not complete, cannot synchronize group members",
+    throw new IntegrationInstanceAuthorizationError(
```
aiwilliams: I'm concerned that assuming the problem was an authorization error could lead to some wild goose chasing. Ideally, this step would be skipped by the state machine when the corresponding fetch state returns certain information into the state data. For now, it would be sufficient to communicate between the fetch state and this one something like `groupsState.authorizationForbidden`, and then handle this by doing nothing.

src/azure/AzureClient.test.ts
```diff
@@ -118,7 +118,11 @@ describe("fetchUsers", () => {
     const warn = jest.spyOn(logger, "warn");
 
     const response = await client.fetchUsers();
-    expect(response).toBeUndefined();
+    expect(response).toEqual({
+      nextLink: undefined,
+      resources: [],
+      err: expect.any(Error),
+    });
```
aiwilliams: I like this change a lot. It's better than `undefined`, allowing the caller too check to see what the error was 👍

src/azure/fetchBatchOfGroupMembers.ts
```diff
@@ -92,13 +93,13 @@ export default async function fetchBatchOfGroupMembers(
     typeof nextLink !== "string" && groupIndex === totalGroups;
 
   await groupsCache.putState({
-    groupMembersFetchCompleted:
-      !fetchErrorOccurred && groupMembersFetchCompleted,
+    groupMembersFetchCompleted: !fetchError && groupMembersFetchCompleted,
+    fetchError: fetchError,
```
aiwilliams: Ahh, nice.

PR #59, opened 2020-02-17
Do not page when access to account org is denied

fomentia:
User configuration error. We should still create the account entity and put it in the cache, but inform the user in the job log that it will be somewhat anemic.

### DIFF
--- a/src/converters/graph.test.ts
+++ b/src/converters/graph.test.ts
@@ -5,9 +5,9 @@ import {
 import { Organization } from "@microsoft/microsoft-graph-types";
 
 import {
-  createAccountEntity,
   createGroupEntity,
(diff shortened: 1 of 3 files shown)


src/converters/graph.test.ts
```diff
@@ -30,7 +30,10 @@ describe("createAccountEntity", () => {
         },
       ],
     };
-    const accountEntity = createAccountEntity(instance, organization);
+    const accountEntity = createAccountEntityWithOrganization(
```
aiwilliams: `describe` description does not match now.

fomentia: @aiwilliams will fix in my other PR

PR #60, opened 2020-02-17
Vm images disks

erkangz:


fomentia:
I think @aiwilliams is gonna have to review this one, since he was more involved and it's pretty large.

erkangz:
> Thanks so much for moving the Azure integration forward @erkangz ! I know that TypeScript hasn't been friendly toward the effort 😬
> 
> I'm bummed this is not using `createIntegrationEntity` for all the new entities. I understand that it would require more effort to get the data model schemas into the SDK. Are there other reasons to avoid `createIntegrationEntity`? I'd like to see us work toward making it what it needs to be.

yeah we'll need to add more to the SDK

PR #67, opened 2020-03-24
Fix storage account oldData CPU usage

fomentia:


fomentia:
Seems to work in dev.

PR #69, opened 2020-03-29
Add support for LB entities and relationships

erkangz:
image

### DIFF
--- a/src/synchronizers/syncronizeComputeResources.ts
+++ b/src/synchronizers/syncronizeComputeResources.ts
@@ -39,6 +39,7 @@ import {
   createDiskEntity,
   createVirtualMachineDiskRelationships,
   createLoadBalancerBackendNicRelationship,
+  createLoadBalancerEntity,
 } from "../converters";
 import {
   AccountEntity,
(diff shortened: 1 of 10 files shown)


src/synchronizers/syncronizeComputeResources.ts
```diff
@@ -409,6 +434,19 @@ async function fetchVirtualMachineDisks(
...
+async function fetchLoadBalaners(
+  client: ResourceManagerClient,
+  webLinker: AzureWebLinker,
+): Promise<EntityFromIntegration[]> {
+  const entities: EntityFromIntegration[] = [];
+  await client.iterateLoadBalancers(e => {
+    entities.push(createLoadBalancerEntity(webLinker, e));
+  });
+  console.log("GETTING LOAD BALANCERS");
+  console.log({ entities });
```
aiwilliams: I think you'll want to remove these log statements.

src/synchronizers/syncronizeComputeResources.ts
```diff
@@ -443,3 +481,30 @@ function findNetworkInterfacesForVM(
...
+): IntegrationRelationship[] {
+  const relationships: IntegrationRelationship[] = [];
+  const lbData = getRawData(lb) as LoadBalancer;
+  if (lbData.backendAddressPools) {
+    lbData.backendAddressPools.forEach(backend => {
+      if (backend.backendIPConfigurations) {
+        backend.backendIPConfigurations.forEach(ip => {
+          if (ip.id) {
+            /**
+             * Need to remove the extra `/ipConfigurations/*` path from the nicId.
```
aiwilliams: I think it would be helpful to indicate the reason this needs to happen.

src/synchronizers/syncronizeComputeResources.ts
erkangz: oops

src/synchronizers/syncronizeComputeResources.ts
erkangz: updated here:
https://github.com/JupiterOne/graph-azure/pull/70/files

PR #71, opened 2020-04-03
Handle 429 failure

aiwilliams:
Please consider the approach documented in the code. I was able to run the new test to work all this out, but I could not keep it enabled because replaying would take over 5 minutes.

### DIFF
--- a/src/azure/resource-manager/ResourceManagerClient.ts
+++ b/src/azure/resource-manager/ResourceManagerClient.ts
@@ -79,13 +79,16 @@ interface ResourceListResponse<T> extends Array<T> {
  * > client-side self-throttling when the available call count for a target
  * > operation group drops below some low threshold.
  *
- * It seems they've coded the ms-rest-js `ServiceClient` in such a way as to
- * enforce that because, once a 429 response is returned, it may be that the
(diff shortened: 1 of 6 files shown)


src/azure/resource-manager/ResourceManagerClient.ts
```diff
@@ -317,17 +403,23 @@ export default class ResourceManagerClient {
...
    * @param rmModule a module that supports listAll() and listAllNext()
    * @param callback a function to receive each resource throughout pagination
+   * @param resourceProviderRateLimit number of requests allowed per rate period
+   * @param resourceProviderRatePeriod number of milliseconds over which rate
+   * limit applies
    */
   private async iterateAllResources<T, L extends ResourceListResponse<T>>(
     rmModule: ResourceManagementModule,
     callback: (r: T) => void | Promise<void>,
+    resourceProviderRatePeriod = 6 * 60 * 1000,
```
aiwilliams, line 413: Note that this 5 minutes is true for some resource provider endpoints, but perhaps not for all of them. It allows for each caller of `iterateAllResources` to provide specific rate periods should that be necessary.

fomentia:
:heart: the documentation!!

PR #73, opened 2020-04-07
SQL database transparent data encryption

aiwilliams:
This includes some refactoring to make the service client available to the callback during iteration of resource lists, so that for an item in a list, additional information can be retrieved.


src/synchronizers/syncronizeDatabaseResources.ts
```diff
@@ -59,100 +55,149 @@ async function synchronize(
...
+
+            if (status) {
+              encrypted = ENCRYPTION_ENABLED_PATTERN.test(status);
+            }
+          } catch (err) {
+            logger.warn(
+              { err, server: server.id, database: database.id },
+              "Failed to obtain transparentDataEncryptions for database",
+            );
+          }
```
aiwilliams, line 189: This is the code that determines whether a database is encrypted. For some reason I have yet to determine, the response comes back without being deserialized/mapped to the type advertised in the TypeScript typings and the documentation. I've kept the code in a way to handle either case, but we'll need to get to the bottom of this.

I wondered if it has to do with forcing a single version with:
```
  "resolutions": {
    "@azure/ms-rest-azure-js": "^2.0.0",
    "@azure/ms-rest-js": "^2.0.0"
  },
```

However, `@azure/arm-sql` uses those versions:
```
    "@azure/ms-rest-azure-js": "^2.0.1",
    "@azure/ms-rest-js": "^2.0.4",
```

I could not locate any recent GitHub Issues for deserialization problems.

Anyhow, I wanted to get this code out and make progress on other `encrypted` fields, so hopefully I'll learn more in time to get this resolved.

src/synchronizers/syncronizeDatabaseResources.ts
```diff
@@ -59,100 +55,149 @@ async function synchronize(
...
+
+          entities.push(
+            createDatabaseEntity(
+              webLinker,
+              database,
+              databaseEntityType,
+              encrypted,
```
aiwilliams, line 196: I'd like to include the Transparent Data Encryption content as part of the raw data, though it currently is nothing more than `status: "Enabled"`.

terraform/storage.tf
```diff
@@ -0,0 +1,40 @@
...
+resource "azurerm_sql_database" "j1dev" {
+  name                = "j1dev-sqldatabase"
+  resource_group_name = azurerm_resource_group.j1dev.name
+  location            = "eastus"
+  server_name         = azurerm_sql_server.j1dev.name
+
+  tags = {
+    environment = "${local.j1env}"
+  }
+}
```
aiwilliams, line 40: I think it useful to provide setup code like this to enable others to contribute to an integration, and to allow the primary maintainer to destroy everything between spells of working on an integration and bring it back up quickly when necessary, or in another trial account when that's a thing.

My guess is that we'll want to be able to limit the resources that are created, when focused on a particular set of resources to ingest, to keep costs down and whatnot, which I think we can easily achieve if we are careful to code so that we can adopt the step pattern we're building in https://github.com/JupiterOne/integration-sdk/pull/2. In the case of Terraform resources, we can conditionally create them by associating them with a step, and then use a var to say which steps we want resources for.

cc/ @charlieduong94 @fomentia @philidem

terraform/storage.tf
```diff
@@ -0,0 +1,40 @@
...
+  storage_account_name = azurerm_storage_account.j1dev.name
+}
+
+resource "azurerm_sql_server" "j1dev" {
+  name                         = "j1dev-sqlserver"
+  resource_group_name          = azurerm_resource_group.j1dev.name
+  location                     = "eastus"
+  version                      = "12.0"
+  administrator_login          = "4dm1n157r470r"
+  administrator_login_password = "4-v3ry-53cr37-p455w0rd"
```
aiwilliams, line 24: This wouldn't be good under normal circumstances of course. In fact, I think I need to figure out how to remove this because once it's merged into master, some hacker is going to come along and make it look like an important problem.

terraform/storage.tf
```diff
@@ -0,0 +1,49 @@
...
+  storage_account_name = azurerm_storage_account.j1dev.name
+}
+
+resource "azurerm_sql_server" "j1dev" {
+  name                         = "j1dev-sqlserver"
+  resource_group_name          = azurerm_resource_group.j1dev.name
+  location                     = "eastus"
+  version                      = "12.0"
+  administrator_login          = random_string.administrator_login.result
+  administrator_login_password = random_password.administrator_password.result
```
aiwilliams, line 24: Updated to use random value, since we will not be accessing the database.

PR #75, opened 2020-04-13
Handle list MySQL databases 502 response

aiwilliams:
I don't love this solution because it can lead to deleting databases during a transient listing error. However, we're seeing a consistent `502` response ("The issue encountered for 'Microsoft.DBforMySQL'; cannot fulfill the request.") for a specific subscription/server, otherwise no failure to list databases.


docs/development.md
```diff
@@ -77,6 +78,19 @@ ARM_ENVIRONMENT=public
...
+### Using `tfenv` on Mac
+
+```sh
+$ brew install tfenv
+$ cat .terraform-version
+$ tfenv install <version>
+$ cd terraform
+$ env $(grep -v '^#' .env) terraform init
+$ env $(grep -v '^#' .env) terraform plan
+```
```
aiwilliams, line 90: I wanted my editor to support the latest TF syntax, which meant I needed a local language server that would run fast on my Mac - no Docker start up costs.

src/azure/resource-manager/ResourceManagerClient.test.ts
```diff
@@ -122,13 +126,131 @@ describe("iteratePublicIPAddresses", () => {
...
+          },
+        });
+      });
+
+    const client = new ResourceManagerClient(config, createTestLogger(), true);
+
+    const iteratee = jest.fn();
+    await expect(
+      client.iterateMySqlDatabases(server, iteratee),
+    ).rejects.toThrow();
```
aiwilliams, line 243: I wanted to test and document the handling of the `502` as an error that the synchronizer must handle.

terraform/storage.tf
```diff
@@ -39,8 +39,46 @@ resource "azurerm_sql_database" "j1dev" {
...
+  # Unsupported by resource at this time 😢
+  # tags = {
+  #   environment = "${local.j1env}"
+  # }
+}
+
 resource "random_string" "administrator_login" {
   length = 13
+  special = true
+  override_special = "_"
```
aiwilliams, line 81: This allowed using the login with MySQL.

PR #77, opened 2020-04-14
Ingest Cosmos DB SQL databases

aiwilliams:
Lots of refactoring, working toward clearer code.

### DIFF
--- a/docs/jupiterone-io/graph-azure.md
+++ b/docs/jupiterone-io/graph-azure.md
@@ -69,52 +69,55 @@ though the `Account` entity will always be ingested.
 | Group Member                    | `azure_group_member` | `User`                |
 | User                            | `azure_user`         | `User`                |
 
-| Azure Resources   | \_type of the Entity        | \_class of the Entity           |
-| ----------------- | --------------------------- | ------------------------------- |
-| Compute           | `azure_vm`                  | `Host`                          |
-|                   | `azure_image`               | `Image`                         |
-|                   | `azure_managed_disk`        | `DataStore`, `Disk`             |
-| Load Balancer     | `azure_lb`                  | `Gateway`                       |
-| Virtual Network   | `azure_vnet`                | `Network`                       |
-| Subnet            | `azure_subnet`              | `Network`                       |
-| Security Group    | `azure_security_group`      | `Firewall`                      |
-| Network Interface | `azure_nic`                 | `NetworkInterface`              |
-| Public IP Address | `azure_public_ip`           | `IpAddress`                     |
-| Blob (Storage)    | `azure_storage_container`   | `DataStore`                     |
-| Databases         | `azure_mariadb_database`    | `Database`, `DataStore`         |
-|                   | `azure_mariadb_server`      | `Database`, `DataStore`, `Host` |
-|                   | `azure_mysql_database`      | `Database`, `DataStore`         |
-|                   | `azure_mysql_server`        | `Database`, `DataStore`, `Host` |
-|                   | `azure_postgresql_database` | `Database`, `DataStore`         |
-|                   | `azure_postgresql_server`   | `Database`, `DataStore`, `Host` |
-|                   | `azure_sql_database`        | `Database`, `DataStore`         |
-|                   | `azure_sql_server`          | `Database`, `DataStore`, `Host` |
+| Azure Resources   | \_type of the Entity          | \_class of the Entity           |
+| ----------------- | ----------------------------- | ------------------------------- |
+| Compute           | `azure_vm`                    | `Host`                          |
+|                   | `azure_image`                 | `Image`                         |
+|                   | `azure_managed_disk`          | `DataStore`, `Disk`             |
+| Load Balancer     | `azure_lb`                    | `Gateway`                       |
+| Virtual Network   | `azure_vnet`                  | `Network`                       |
+| Subnet            | `azure_subnet`                | `Network`                       |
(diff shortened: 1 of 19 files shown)


docs/jupiterone-io/graph-azure.md
```diff
@@ -69,52 +69,55 @@ though the `Account` entity will always be ingested.
...
+| Blob (Storage)    | `azure_storage_container`     | `DataStore`                     |
+| Databases         | `azure_mariadb_database`      | `Database`, `DataStore`         |
+|                   | `azure_mariadb_server`        | `Database`, `DataStore`, `Host` |
+|                   | `azure_mysql_database`        | `Database`, `DataStore`         |
+|                   | `azure_mysql_server`          | `Database`, `DataStore`, `Host` |
+|                   | `azure_postgresql_database`   | `Database`, `DataStore`         |
+|                   | `azure_postgresql_server`     | `Database`, `DataStore`, `Host` |
+|                   | `azure_sql_database`          | `Database`, `DataStore`         |
+|                   | `azure_sql_server`            | `Database`, `DataStore`, `Host` |
+| Cosmos DB         | `azure_cosmosdb_account`      | `Account`                       |
```
erkangz, line 92: perhaps this can be `Account` and `Service` dual class? 

Also -- do we create `azure_storage_account` entity?

aiwilliams: > Also -- do we create `azure_storage_account` entity?

https://github.com/JupiterOne/graph-azure/blob/JPT-3539-ingest-cosmosdb/src/jupiterone/entities.ts#L145 indicates there is no `azure_storage_account` entity.

There currently is no `azure_storage_account` entity, only `azure_storage_blob_service` etc. (an entity per service). When a Storage Account has no activated storage services, the integration logs `"Storage account has no registered service endpoints, nothing to synchronize"`.

I'm thinking there needs to be a Storage Account entity because there can be more than one, so we should contain the `azure_storage_blob_service` etc. entities within their Storage Account, but more importantly, Storage Accounts have settings - as do Cosmos DB Accounts - that include location, access control, and other settings we'll need to ingest.

> perhaps this can be `Account` and `Service` dual class?

I think for Cosmos DB that makes good sense 👍 For Storage, **what do you think** about having the `azure_storage_account` as `_class: "Account"`, and keep the existing `azure_storage_blob_service` as only `_class: "Service"`. 

---

Storage is structured in Azure:

1. There are Storage Accounts, which
2. Have various services (Blob, File, Queue, Table), that
3. Have concepts of containment

And Cosmos DB:

1. There are Cosmos DB Accounts, which
2. Contain N databases (of a specific type (SQL, MongoDB, Gremlin, Cassandra), that
3. Have concepts of containment

erkangz: thanks for the details, @aiwilliams. Looks good to me!

terraform/cosmosdb.tf
```diff
@@ -0,0 +1,38 @@
...
+
+  name                = "j1dev"
+  resource_group_name = azurerm_resource_group.j1dev.name
+  account_name        = azurerm_cosmosdb_account.j1dev.name
+  throughput          = 400
+
+  # Unsupported by the resource
+  # tags = {
+  #   environment = local.j1env
+  # }
```
aiwilliams, line 37: The Cosmos DB databases do not have tagging support, though the Cosmos DB account does. I think the approach for classification is going to be that users should limit a storage account to a class of data. @erkangz I think our packaged question for this integration will need to asses the database in context of the account.

aiwilliams: @erkangz, what do you think about transferring tags from the DB account to the databases?

src/synchronizers/synchronizeCosmosDBAccounts/converters.test.ts
```diff
@@ -0,0 +1,160 @@
...
+  // publicNetworkAccess: "Enabled",
+  enableAutomaticFailover: false,
+  enableMultipleWriteLocations: false,
+  // enablePartitionKeyMonitor: false,
+  isVirtualNetworkFilterEnabled: false,
+  virtualNetworkRules: [],
+  // EnabledApiTypes: "Sql",
+  disableKeyBasedMetadataWriteAccess: false,
+  // enableFreeTier: false,
+  // apiProperties: null,
```
aiwilliams, line 34: These are commented out because the TS types don't support them, but they do come back in the data. I think they're going to be important to ingest, but I'm going to move on for now. These tests help record what is included/transferred.

I think some of these other properties would be a waste of resources to add to the graph object.

src/synchronizers/synchronizeCosmosDBAccounts/converters.test.ts
```diff
@@ -0,0 +1,160 @@
...
+        "/subscriptions/dccea45f-7035-4a17-8731-1fd46aaa74a0/resourceGroups/j1dev/providers/Microsoft.DocumentDB/databaseAccounts/j1dev",
+      displayName: "j1dev",
+      encrypted: true,
+      name: "j1dev",
+      region: "East US",
+      resourceGroup: "j1dev",
+      "tag.environment": "j1dev",
+      "tag.classification": "production",
+      environment: "j1dev",
+      classification: "production",
```
aiwilliams, line 155: I decided to go ahead and transfer tags from the account to the sql databases since they are so connected to the configuration of the account, and do not have their own tags.

PR #78, opened 2020-04-16
Add Storage account file shares

aiwilliams:
I did a good bit of refactoring to support Tables, but abandoned those for now. This allowed me to keep the refactoring and add value by ingesting File Shares.


jest.config.js
```diff
@@ -1,5 +1,7 @@
 const { defaults } = require("jest-config");
 
+process.env.ENABLE_GRAPH_OBJECT_SCHEMA_VALIDATION = 1;
```
aiwilliams, line 3: Now that we have the complete set of schemas, we should enable test-time and local execution-time schema validation. It will help us drive toward greater value in the `@jupiterone/data-model`.

cc/ @charlieduong94

charlieduong94: :100:

PR #90, opened 2020-05-05
Fix group iteration

aiwilliams:


aiwilliams:
GitHub, why you no run my Actions!?!

PR #93, opened 2020-05-12
Process db servers independently

aiwilliams:
Some Azure API calls are logged, but then the step goes quiet and eventually times out. I think we might be processing too much data all at once. We never get past the MySQL databases, so I think we need to break it down further and process one server and it's databases at a time.

### DIFF
--- a/src/synchronizers/syncronizeDatabaseResources.ts
+++ b/src/synchronizers/syncronizeDatabaseResources.ts
@@ -63,15 +63,30 @@ async function synchronize(
 
   logger.info({ dbType }, "Synchronizing databases...");
 
+  const results: PersisterOperationsResult[] = [];
+
   const newServerEntities = await fetchDbServers(
     executionContext,
     webLinker,
     dbType,
   );
 
-  const newDatabaseEntities: EntityFromIntegration[] = [];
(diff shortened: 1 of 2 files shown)


src/synchronizers/syncronizeDatabaseResources.ts
```diff
@@ -96,37 +109,61 @@ async function synchronize(
...
+  // );
 
-  const oldServerDbRelationships = await graph.findRelationshipsByType(
-    serverDatabaseRelationshipType,
-  );
+  // results.push(
+  //   await persister.publishEntityOperations(
+  //     persister.processEntities(orphanedDatabaseEntities, []),
+  //   ),
+  // );
```
aiwilliams, line 154: The problem is, we need to synchronize in subsets 😞 So, this change adds a `serverId` property to the databases. However, it will not find databases that do not have that property, so I think we'd need something like this to delete those without it. This would probably need to execute before the rest of the code. What do you think about this @erkangz @philidem ?

erkangz: will this still be an issue if we switch to using the new sdk?

aiwilliams:
Integration has been moved to new integration-sdk. This is no longer needed.

PR #101, opened 2020-05-27
Refactor neo integration

aiwilliams:
Alright, this is a big one. The only goal was to restructure the program to fit in the new integration pattern of small steps, doing nothing more than fetching data. I'd recommend you just `git checkout refactor-neo-integration` in your clone of this repo and browse starting at `src/index.ts`.

@erkangz I'd really appreciate you pulling this, doing a `yarn install` and `yarn j1-integration collect` (which will say you need to fix your `.env` file to drop the prefix on the variables) against your Azure account, which I hope has all the resources you've tested against 😄I have updated this to include recent changes you've made in `master`.


erkangz:
Finally got to it.  Ran it locally and all seemed fine.

PR #104, opened 2020-07-13
Fix invalid entity property _integrationInstanceId

aiwilliams:
Working toward changes to avoid sending the `_integrationInstanceId` property when building a relationship between a security group and a private azure subnet network entity. This refactoring helped me to understand what is going on, which I documented along the way.

@erkangz, is there any reason we *must* use a mapped relationship for rules that reference private subnets? I think we could instead look up already generated `azure_subnet` entities in the `jobState` and create direct references, but I want to be sure I'm not missing something.


src/steps/resource-manager/network/converters/securityGroups.ts
```diff
@@ -18,29 +18,63 @@ import {
...
   properties: FirewallRuleProperties & {
     _type: string;
-    srcPortRange?: string;
-    srcPortRanges?: string;
-    dstPortRange?: string;
-    dstPortRanges?: string;
+    // srcPortRange?: string;
+    // srcPortRanges?: string;
+    // dstPortRange?: string;
+    // dstPortRanges?: string;
```
aiwilliams, line 77: I'm thinking to drop these because I don't see any reference to them. Having them here led me to think about what they mean, only to find they don't seem to mean anything for processing the data. I suspect they're transferred by `convertProperties`, but I don't think all the transferred properties are represented here, which is fine, so these probably don't need to be either.

aiwilliams:
I talked with @philidem and we agreed that it would be best to focus on making changes in the integration, both to avoid special handling in the persister and to clarify the fact that the integration wants to make direct relationships within it's own set of data. I think this is a common pattern where the integration needs a way to find entities it has already created and make direct internal relationships.

src/steps/resource-manager/network/converters/securityGroups.ts
erkangz: we used them at one point.  probably same to drop

erkangz:
> Working toward changes to avoid sending the `_integrationInstanceId` property when building a relationship between a security group and a private azure subnet network entity. This refactoring helped me to understand what is going on, which I documented along the way.
> 
> @erkangz, is there any reason we _must_ use a mapped relationship for rules that reference private subnets? I think we could instead look up already generated `azure_subnet` entities in the `jobState` and create direct references, but I want to be sure I'm not missing something.

If the key can be determined.  The reason for using mapped relationships is because we only know the CIDR

aiwilliams:
This has been changed more to build the relationships after both the networks and the security groups have been fetched, in a later step that depends on those two steps.

All the tests pass, but when I execute the integration locally, there are some schemas that say other entities are invalid 🤔

PR #109, opened 2020-07-17
Update docs for new developers

aiwilliams:
Based on working to get @ndowmon going on development.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -5,10 +5,9 @@ There are a few steps to take to setup Azure integration development:
 1. Identify an Azure account for local development. It is highly recommended
    that you avoid targeting a production Active Directory or Subscription,
    because tests rely on Polly.js recordings for playback in CI/CD environments.
-   JupiterOne staff developers each have their own Azure Subscription.
 1. Create credentials with write permissions to configure
    `<graph-azure>/terraform/.env`. The Terraform Azure provider client needs a
-   lot more permissions than the integration itself
+   lot more permissions than the integration itself.
 1. Create an App Registration with read permissions to configure
    `<graph-azure>/.env`. The integration Azure client will use this for
    ingesting information about the targeted Active Directory and Azure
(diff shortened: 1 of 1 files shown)


docs/development.md
```diff
@@ -62,21 +76,74 @@ Creating a role assignment under the scope of "/subscriptions/dccea..."
...
+```txt
+CLIENT_ID=app_registration_client_id
+CLIENT_SECRET=app_registration_client_secret
+DIRECTORY_ID=development_directory_id
+SUBSCRIPTION_ID=development_subscription_id
+INGEST_ACTIVE_DIRECTORY=true
 ```
 
-## Terraforming for Development
+You should not be able to [execute the integration](#execute-the-integration)
```
ndowmon, line 130: Is this supposed to say "You should _now_ be able to execute the integration"?

docs/development.md
```diff
@@ -62,21 +76,74 @@ Creating a role assignment under the scope of "/subscriptions/dccea..."
...
+with these credentials. Note that [running the tests](#running-tests) requires
+more attention.
+
+## Executing Terraform
+
+Example Azure resources are maintained with Terraform to allow for easy setup
+and teardown, and to avoid unnecessary expenses between development cycles. See
+Terraform code where `count` is used to force a decision about which resources
+to build. Once an integration step has a recording made, those resources can be
+set back to count `0`.
```
ndowmon, line 140: For my own info, when you do want to generate some of these resources, are you simply resetting the `default` values in `terraform/vars.tf`? Or do you override them with environment variables?

docs/development.md
```diff
@@ -62,21 +76,74 @@ Creating a role assignment under the scope of "/subscriptions/dccea..."
...
+Using the Application (client) ID, Directory (tenant) ID, and generated secret
+from the App Registration, create `<graph-azure>/.env` to establish
+authentication values for the integration API clients. This file is already in
+`.gitignore`, and is distinct from the `<graph-azure>/terraform/.env` file read
+by Terraform.
+
+```txt
+CLIENT_ID=app_registration_client_id
+CLIENT_SECRET=app_registration_client_secret
+DIRECTORY_ID=development_directory_id
```
ndowmon, line 125: not sure if it's good to postpend `# same as above` to DIRECTORY_ID and SUBSCRIPTION_ID or `#different from above` to CLIENT_ID and CLIENT_SECRET, these two sets of vars in the same document do tend to look similar and could maybe be confused without a specific callout?

docs/development.md
aiwilliams: Add a tip, that's a great way to do it!

PR #114, opened 2020-07-22
Add iam roles

ndowmon:
I struggled through an issue where custom roles are _not_ being returned from the client call to `roleDefinitions`. Decided to table that and tackle it in a future/smaller PR. 

This adds role definitions which are used by any active role assignments. The next step will be to add the role assignments mapping to either discrete service principals or mapped relationships.


package.json
```diff
@@ -14,7 +14,7 @@
   },
   "scripts": {
     "start": "j1-integration collect",
-    "graph": "j1-integration visualize",
+    "graph": "j1-integration visualize && open .j1-integration/graph/index.html",
```
aiwilliams, line 17: I'm not entirely sure this will work on Windows 🤔 I think it's fine to keep it here, but we will hopefully one day see people in the Windows community help us add stuff to this integration, at which point we may have to drop the platform specific `open`.

src/steps/resource-manager/authorization/client.test.ts
```diff
@@ -0,0 +1,97 @@
...
+      }),
+      expect.objectContaining({
+        id: expect.any(String),
+        name: '41c65a1f-1d8f-450f-b8c5-f56fd3b10302',
+      }),
+      expect.objectContaining({
+        id: expect.any(String),
+        name: '1f0b8507-2f08-467b-8347-f5ca0554d696',
+      }),
+    ]);
```
aiwilliams, line 95: As we work toward a world where we can run this against a different subscription, would it be good to make these `name: expect.stringMatching(niceRegexpForThoseValuesHere)`?

src/steps/resource-manager/authorization/converters.test.ts
```diff
@@ -0,0 +1,52 @@
...
+        '/providers/Microsoft.Authorization/roleDefinitions/8e3af657-a8ff-443c-a75c-2fe8c4bcb635',
+      ),
+      rules: [
+        {
+          actions: ['*'],
+          dataActions: [],
+          notActions: [],
+          notDataActions: [],
+        },
+      ],
```
aiwilliams, line 49: Sadly, we cannot store an array of objects in an entity. 😢 It can store an array of strings though. I think we should code for `actions: string[]`, `dataActions: string[]`, `notActions: string[]`, and `notDataActions: string[]`, and just combine the values from all the objects in the `data.permissions` array. This seems acceptable after reviewing https://docs.microsoft.com/en-us/azure/role-based-access-control/role-definitions, where they just collapse them when displayed in their own tools.

Is `rules` is a required property of AccessPolicy?

package.json
ndowmon: True, and bummer. I hadn't thought of that but I think this should be removed.

src/steps/resource-manager/authorization/client.test.ts
ndowmon: I guess even with that regex, the value of this particular expect statement is basically voided if another developer re-records the HTTP requests for this specific test. All it's basically saying is "there are exactly 5 role assignments returned from `client.iterateRoleAssignments`. So if another dev has 4 or 6 or any number other than 5, this will fail. 

The only response we could technically expect to exist for any dev is the `azurerm_role_assignment.j1dev` resource created in `iam.tf`. I'll update the TF so that we can test/identify the created role assignment, and make this matcher looser while ensuring that role assignment exists.

src/steps/resource-manager/authorization/converters.test.ts
ndowmon: `rules` is an optional property but not required - I was hoping to use it just for consistency across cloud providers (Not sure if AWS actually uses `rules`, but trying to adhere to specs where I can).

Given the following permissions: `[{ actions: ['action-1', 'action-2'], notActions: ['not-action-1', 'not-action-2']}]`...

Do you suggest 
  - `data.properties = ['action = action-1', 'action = action-2', 'not-action = not-action-1', 'not-action = not-action-2']`
  - `data.properties = ['actions = action-1, action-2', 'not-action = not-action-1, not-action-2']`

I'm just trying to think about how users would want most likely want to query this data. The first is most verbose, but would allow users to easily query for something like `where data.rules contains 'action = *`.

aiwilliams: I was thinking this:

```
entity = {
  actions: ['action-1', 'action-2'],
  notActions: ['not-action-1', 'not-action-2'],
  dataActions: [],
  notDataActions: []
}
```

This way folks can do: `find entity with notActions = 'not-action-1'`.

aiwilliams: Regarding the `rules` property, I don't believe anything uses that right now. `find * with rules != undefined LIMIT 1` returns no results in our J1 account. 😬 

The data model has been more of a guideline than a specification historically. @austinkelleher, @charlieduong94  and others, including myself - as seen in the SDK, where the `createIntegrationEntity` works to validate the entities against the data model schema - regularly consider how we might push improvements into the data model so that it can become more and more of a specification, and ultimately be input into backend data store schemas to help with optimizations. Customers certainly desire consistency across the data.

One reason a standard will be difficult is, folks using Azure will expect the RoleDefinition `['Role', 'AccessPolicy']` entity to express terms used by that platform. If they type in the Landing Page `find azure_ro$` (where `$` is the cursor), and the autocomplete presents `azure_role_definition`, then `find azure_role_definition with $` would present the completions `['actions', 'dataActions', 'notActions', 'notDataActions']`. Of course, it could present `rules` as well, but it's difficult to use that array in a way that would be as intuitive as the proposal above.

Hope that helps give some context!

terraform/iam.tf
```diff
@@ -0,0 +1,17 @@
...
+  assignable_scopes = [
+    data.azurerm_subscription.j1dev.id,
+  ]
+  permissions {
+    actions = ["*"]
+  }
+}
+
+resource "azurerm_role_assignment" "j1dev" {
+  name               = "10000000-0000-0000-0000-000000000000"
```
aiwilliams, line 13: Ah, nice, didn't notice before that the assignment can be named 👍

src/steps/resource-manager/authorization/converters.test.ts
```diff
@@ -0,0 +1,56 @@
...
+      type: 'Microsoft.Authorization/roleDefinitions',
+      webLink: webLinker.portalResourceUrl(
+        '/providers/Microsoft.Authorization/roleDefinitions/8e3af657-a8ff-443c-a75c-2fe8c4bcb635',
+      ),
+      rules: [
+        'actions = *',
+        'actions = other-action',
+        'notActions = not-action',
+        'actions = a-third-action',
+      ],
```
aiwilliams, line 53: How do you feel about dropping rules, based on the commentary about `entity.actions`, `entity.notActions`, etc.?

src/steps/resource-manager/authorization/converters.test.ts
ndowmon: @aiwilliams The use case you provided (find entity with notActions = 'not-action-1') makes a ton of sense to me. I agree with your recommendation that j1 entity properties match naming conventions from the target integration.

Let me make this change quick and then request another review from you!

src/steps/resource-manager/authorization/converters.test.ts
ndowmon: Agreed!

PR #115, opened 2020-07-23
Add custom role descriptions

ndowmon:
Swap out `iterateRoleDefinitions` for `getRoleDefinition`, since many built-in roles are irrelevant and `AuthorizationManagementClient.roleDefinitions.list(<scope>)` doesn't necessarily return all roles that are valid from a role assignment.


ndowmon:
Before merging, note that this currently targets the `add-iam-roles` branch. If that is merged first, please re-target `master`.

src/steps/resource-manager/authorization/index.ts
```diff
@@ -49,12 +49,22 @@ export async function fetchRoleDefinitions(
     ),
   );
 
-  await client.iterateRoleDefinitions(async (rd) => {
-    if (roleDefinitionsUsedInAssignments.has(rd.id as string)) {
-      const roleDefinitionEntity = createRoleDefinitionEntity(webLinker, rd);
+  for (const roleDefinitionId of roleDefinitionsUsedInAssignments) {
+    console.log(roleDefinitionId);
```
aiwilliams, line 53:
```suggestion
```

PR #117, opened 2020-07-24
Add mapped relationships for Role Assignments

ndowmon:


src/steps/resource-manager/authorization/index.ts
```diff
@@ -66,6 +72,27 @@ export async function fetchRoleDefinitions(
...
+  executionContext: IntegrationStepContext,
+): Promise<void> {
+  const { jobState } = executionContext;
+  const accountEntity = await jobState.getData<Entity>(ACCOUNT_ENTITY_TYPE);
+
+  const webLinker = createAzureWebLinker(accountEntity.defaultDomain as string);
+
+  const roleAssignments = await jobState.getData<RoleAssignment[]>(
+    ROLE_ASSIGNMENT_DATA_KEY,
+  );
```
aiwilliams, line 85: I think it would be good to build the relationships and put them in the `jobState` [during `fetchRoleAssignments`](https://github.com/JupiterOne/graph-azure/blob/399a990e7b83b7fcd015b78100bb1eeb35e1cdbe/src/steps/resource-manager/authorization/index.ts#L27). This avoids collecting all the raw data in memory, the key intention of the [pattern implemented by `iterateRoleAssignments`](https://github.com/JupiterOne/graph-azure/blob/399a990e7b83b7fcd015b78100bb1eeb35e1cdbe/src/steps/resource-manager/authorization/client.ts#L37) and other such API client functions.

Each step should work to push entities or relationships into the `jobState`, and later steps may fetch those entities from the `jobData` (which will include `_rawData`, readable with `getRawData(entity)`). The main intention of `jobState.setData`/`jobState.getData` is to allow for transmitting meta data between the steps, or perhaps a singleton entity such as the `Account`.

src/steps/resource-manager/authorization/converters.ts
```diff
@@ -35,3 +48,33 @@ export function createRoleDefinitionEntity(
...
+      ...convertProperties(roleAssignment),
+      webLink: webLinker.portalResourceUrl(roleAssignment.id),
+    },
+    _mapping: {
+      relationshipDirection: RelationshipDirection.FORWARD,
+      sourceEntityKey: roleDefinitionKey,
+      targetFilterKeys: ['id'],
+      targetEntity: {
+        id: roleAssignment.principalId,
+      },
```
aiwilliams, line 76:
```suggestion
      targetFilterKeys: [['_type', 'id']],
      targetEntity: {
        _type: targetEntityType,
        id: roleAssignment.principalId,
      },
```

I think this should include the `_type` of the entity, such as `azure_user` or `azure_group` or `azure_application` etc. This will ensure that when the mapper creates the target entity, it will have `_type` and `id` on it. In fact, in cases where we know how the `_key` is constructed, we should have all of `_type`, `_key`, and `id` on the target.

This will ensure the mapper relates to the existing `azure_user`, for example.

src/steps/resource-manager/authorization/converters.ts
```diff
@@ -35,3 +48,33 @@ export function createRoleDefinitionEntity(
...
+  const targetType = getJupiterTypeForPrincipalType(
+    roleAssignment.principalType as PrincipalType,
+  );
+  const roleDefinitionKey = (roleAssignment.roleDefinitionId as string).replace(
+    roleAssignment.scope as string,
+    '',
+  );
+  return createIntegrationRelationship({
+    _class: ROLE_ASSIGNMENT_RELATIONSHIP_CLASS,
+    _type: createRoleAssignmentRelationshipType(targetType),
```
aiwilliams, line 65: Generally, I think it [best to avoid specifying the relationship `_type` for mapped relationships](https://github.com/JupiterOne/sdk/blob/c1d901fbaaa5b747482471a1b8478b04af37de85/packages/integration-sdk-core/src/data/createIntegrationRelationship.ts#L104), since the function will build a good value based on the source and target values. I note below that the `targetEntity._type` should be specified.

src/steps/resource-manager/authorization/constants.ts
```diff
@@ -8,3 +17,102 @@ export const STEP_RM_AUTHORIZATION_ROLE_DEFINITIONS =
...
+  'rm-authorization-role-assignment-relationships';
+export const ROLE_ASSIGNMENT_RELATIONSHIP_CLASS = 'TRUSTS';
+
+interface RoleAssignmentMap {
+  principalType: PrincipalType;
+  type: string;
+  dependsOn: string[];
+}
+
+export const ROLE_ASSIGNMENT_TYPES_MAP: RoleAssignmentMap[] = [
```
aiwilliams, line 32: I think this map can be simplified, and some of the functions below it removed, if other suggestions are adopted, since we only need to know what the `targetEntity._type` is for a particular `principalType` (`{ 'Application': 'azure_application', 'Group': 'azure_group', ... }`), and also renamed to `ROLE_ASSIGNMENT_PRINCIPAL_ENTITY_TYPE_MAP`.

src/steps/resource-manager/authorization/constants.ts
```diff
@@ -8,3 +17,102 @@ export const STEP_RM_AUTHORIZATION_ROLE_DEFINITIONS =
   'rm-authorization-role-definitions';
 export const ROLE_DEFINITION_ENTITY_TYPE = 'azure_role_definition';
 export const ROLE_DEFINITION_ENTITY_CLASS = ['AccessRole', 'AccessPolicy'];
+
+// Create Role Assignment Relationships
+export const STEP_RM_AUTHORIZATION_ROLE_ASSIGNMENT_RELATIONSHIPS =
+  'rm-authorization-role-assignment-relationships';
+export const ROLE_ASSIGNMENT_RELATIONSHIP_CLASS = 'TRUSTS';
```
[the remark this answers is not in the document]
ndowmon: I'm definitely good with that. I had seen the aws integration docs you mentioned re: `aws_iam_role --TRUSTS--> aws_iam_user`, but I can see how that use case is different! Switched to `ASSIGNED`

src/steps/resource-manager/authorization/index.ts
```diff
@@ -81,4 +108,15 @@ export const authorizationSteps = [
     dependsOn: [STEP_AD_ACCOUNT, STEP_RM_AUTHORIZATION_ROLE_ASSIGNMENTS],
     executionHandler: fetchRoleDefinitions,
   },
+  {
+    id: STEP_RM_AUTHORIZATION_ROLE_ASSIGNMENT_RELATIONSHIPS,
+    name: 'Role Assignment Relationships',
+    types: ROLE_ASSIGNMENT_RELATIONSHIP_TYPES,
+    dependsOn: [
+      STEP_AD_ACCOUNT,
+      STEP_RM_AUTHORIZATION_ROLE_DEFINITIONS,
```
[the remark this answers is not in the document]
ndowmon: I actually had intended it to be `ROLE_DEFINITIONS` as written, because I was planning to find the actual `azure_role_definition` entity and use it as the `MappedRelationship.source` using the `MappedRelationshipOptions` [interface](https://github.com/JupiterOne/sdk/blob/master/packages/integration-sdk-core/src/data/createIntegrationRelationship.ts#L33)

I ended up instead using the `MappedRelationshipLiteralOptions` [interface](https://github.com/JupiterOne/sdk/blob/master/packages/integration-sdk-core/src/data/createIntegrationRelationship.ts#L86) whose `_mapping` property takes just a `sourceEntityKey`. 

I am more interested in switching to use the `MappedRelationshipOptions` actually since it's a safer way to ensure that the source entity does exist. Let's have a quick chat about this this AM.

src/steps/resource-manager/authorization/converters.ts
ndowmon: Totally agree

ndowmon: Adding `_key` and `_type` would allow us to use the `MappedRelationshipOptions` interface for `createRelationship()`, which is a safer option than the `MappedRelationshipLiteralOptions` interface.

src/steps/resource-manager/authorization/converters.ts
ndowmon: Noted. I think I saw other parts of this integration explicitly calling `generateRelationshipType()` but if the SDK takes care of that I will leave it to the SDK!

ndowmon: One question - without this, i'm leaving the step's `types` field as an empty list in `index.ts`

Is that a problem?

aiwilliams: > I think I saw other parts of this integration explicitly calling `generateRelationshipType()` but if the SDK takes care of that I will leave it to the SDK!

The function mainly serves to build the `_type` value when it needs to be known outside the `createIntegrationRelationship` function. That is, if something needs to later look up the relationships by `_type`, it can use this function *just as `createIntegrationRelationship` does*, to get the same value. Does that makes sense?

aiwilliams: > One question - without this, i'm leaving the step's `types` field as an empty list in `index.ts`
> 
> Is that a problem?

Ah, right, so I've done a constant like `ROLE_ASSIGNMENT_RELATIONSHIP_TYPE = generateRelationshipType(...)`, and don't assign it in `createIntegrationRelationship`, but use `ROLE_ASSIGNMENT_RELATIONSHIP_TYPE` in the `types` field.

Perhaps it would be clearer to guide folks to defining those constants and *require* `_type` in the `createIntegrationRelationship` function? 🤔

ndowmon: You might be right about that. If devs still need to explicitly give the relationship `_type` in an integration step, maybe they should be the ones in charge of creating/handling the `_type`.

That said, I'm not sure I 100% understand the logic behind requiring `types` in an integration step. Is it used for a sanity check? On one hand, I will say, it was helpful that `j1-integration collect` gave me a warning that the type `mapping_source_assigned_azure_user` was not defined in `types`. On the other hand, I don't think that types really _ought_ to be required for relationships - the purpose that relationship types serve in our system seems almost negligible.

src/steps/resource-manager/authorization/constants.ts
ndowmon: I'm still hoping to maintain this map to `dependsOn` _and_ `type`, especially since we're now using `getEntity` and direct relationships whenever possible. But maybe there's no value in defining `dependsOn` in that same map, since all the `dependsOn` are just flattened to a 1d array anyway.

ndowmon: For now I'd like to keep the implementation as is. I fear that for some of the other principal types, we can't rely on just one step (e.g. maybe the `PrincipalType` of `Application` ends up pulling from multiple RM APIs for different types of applications... it will be nice to logically group any associated steps, types & `PrincipalType`s. 

This may be a case of YAGNI but I wouldn't call it _extremely_ overdesigned/complex at the moment, just a little bit.

src/steps/resource-manager/authorization/converters.ts
aiwilliams: Take a look at https://github.com/JupiterOne/sdk/blob/master/docs/integrations/development.md#letting-the-synchronizer-know-about-partial-datasets and see if that explains the need for `types`. TLDR; it's a critical component of the sync mechanisms in the persister.

src/steps/resource-manager/authorization/constants.ts
aiwilliams: Sounds good, thanks for considering @ndowmon 👍  It does serve a purpose now that the source entity is being looked up.

src/steps/resource-manager/authorization/converters.test.ts
```diff
@@ -36,7 +43,7 @@ describe('createRoleDefinitionEntity', () => {
     expect(createRoleDefinitionEntity(webLinker, data)).toEqual({
       ...convertProperties(data),
       _key:
-        '/providers/Microsoft.Authorization/roleDefinitions/8e3af657-a8ff-443c-a75c-2fe8c4bcb635',
+        'azure_role_definition_/providers/Microsoft.Authorization/roleDefinitions/8e3af657-a8ff-443c-a75c-2fe8c4bcb635',
```
aiwilliams, line 46: I'm a bit torn on the practice of including `_type` in the `_key`, because @philidem and I have talked here and there about `_key` being unique within the `scope` of an integration (`_integrationInstanceId`). We should be able to change the `_type` of something without destroying the history of the object, for example. Perhaps we should have a discussion about this and resolve once-for-all our going forward position on this.

Of course, there may be a need for something like a prefix in some integrations because the natural `id` property of the entity is not unique. That is not a concern in Azure since the `id` includes type information and uuid values.

cc/ @austinkelleher

src/steps/resource-manager/authorization/converters.test.ts
```diff
@@ -52,3 +59,120 @@ describe('createRoleDefinitionEntity', () => {
...
+      _key: 'azure_role_definition-1|assigned|azure_user-1',
+      _toEntityKey: 'azure_user-1',
+      _type: 'azure_role_definition_assigned_user',
+      displayName: 'ASSIGNED',
+      id:
+        '/subscriptions/d3803fd6-2ba4-4286-80aa-f3d613ad59a7/providers/Microsoft.Authorization/roleAssignments/c967042a-aad6-4e3b-8485-1c85d5e6f9e8',
+      name: 'c967042a-aad6-4e3b-8485-1c85d5e6f9e8',
+      principalId: 'b6e7f627-8731-47bd-be0e-477d1cbc6e17',
+      principalType: 'User',
+      roleDefinitionId:
```
aiwilliams, line 108: Excellent, glad to see these on the relationships.

src/steps/resource-manager/authorization/index.ts
```diff
@@ -39,46 +50,87 @@ export async function fetchRoleDefinitions(
...
-
-  const roleDefinitionsUsedInAssignments = new Set(
-    roleAssignments.map(
-      (ra) => ra.roleDefinitionId?.replace(ra.scope as string, '') as string,
-    ),
-  );
+  const roleAssignments: RoleAssignment[] = [];
+  await client.iterateRoleAssignments(async (ra) => {
+    roleAssignments.push(ra);
+  });
```
aiwilliams, line 56: How do you feel about processing each assignment within this loop, instead of pulling it all into memory, then iterating through the list two more times? For example, something along the lines:

```ts
await client.iterateRoleAssignments(async (roleAssignment) => {
  const [roleDefinitionEntity, targetEntity] = await Promise.all([
    findOrCreateRoleDefinitionEntity(roleAssignment),
    findOrBuildTargetEntityForRoleDefinition(roleAssignment)
  ]);

  const relationshipProperties = getRoleAssignmentRelationshipProperties(
    webLinker,
    roleAssignment,
  );

  if (targetEntity._rawData) { // I don't love this, how can we otherwise know 🤔
        createDirectRelationship({
          _class: ROLE_ASSIGNMENT_RELATIONSHIP_CLASS,
          from: roleDefinitionEntity,
          to: targetEntity,
          properties: relationshipProperties
        }),
      );
  } else {
        createMappedRelationship({
          _class: ROLE_ASSIGNMENT_RELATIONSHIP_CLASS,
          source: roleDefinitionEntity,
          target: { _type: targetType, _key: targetKey },
          properties: relationshipProperties
        }),
      );
  }
});
```

src/steps/resource-manager/authorization/converters.ts
ndowmon: Ok. Makes sense. So the `_type` had better be consistent over sequential integration runs, but for j1ql, searching, etc (everyday customer use) the exact _value_ of the `_type` is not terribly important. 

In this case, whether it's `mapping_source_assigned_azure_user` or `azure_role_definition_assigned_user` will not make much of a difference from the user's perspective, _or_ the synchronizer's behavior. But for the sake of underlying data clarity it's good to have the latter.

src/steps/resource-manager/authorization/converters.test.ts
ndowmon: I made this change because it looks like the `step-users` step creates keys prefixed with `_type`, which I needed to leverage in order to properly link `azure_role_definition`s to `azure_user`s. It looks like the remainder of the integration uses type-prefixed `key`s, so I wanted to stay consistent...

ndowmon: Maybe we should just move forward without prefixing `_type` on future `_key`s, or maybe we should pick up some tech debt work to go back and remove `_type` from all existing `_key`s as well.

src/steps/resource-manager/authorization/index.ts
ndowmon: I like that! Heading out for haircut now but let me look at reimplementing once I get back.

src/steps/resource-manager/authorization/converters.ts
aiwilliams: > for the sake of underlying data clarity it's good to have [`azure_role_definition_assigned_user`].

Definitely much clearer, that was a great change!

src/steps/resource-manager/authorization/converters.test.ts
aiwilliams: I see. How about for now, since it's just you and I in here 😄 , we move forward without prefixing `_type` on future `_key`s, and as we can here and there, we consider the implications of fixing the prefixed `_type` as we run into an entity like that for the next relationship? That will move us toward getting more ingestion released, and we can pick up the debt when you get back from your break.

src/steps/resource-manager/authorization/index.ts
aiwilliams: Okay, great! I think of this as a "streaming" approach, which drives a pattern that should make it easier to process events, such as role assignments reported through a something like CloudTrail (don't know the Azure equivalent).

src/steps/resource-manager/authorization/converters.test.ts
ndowmon: Sounds good to me - the recent change removes the `_type` prefix from `azure_role_definition` entities.

src/steps/resource-manager/authorization/index.ts
ndowmon: Thanks for the great suggestion. Implemented in the new change!

src/steps/resource-manager/authorization/index.ts
```diff
@@ -1,36 +1,131 @@
...
+  const roleDefinitionKey = getRoleDefinitionKeyFromRoleAssignment(
+    roleAssignment,
+  );
+  let roleDefinitionEntity: Entity;
+  try {
+    roleDefinitionEntity = await jobState.getEntity({
+      _type: ROLE_DEFINITION_ENTITY_TYPE,
+      _key: roleDefinitionKey,
+    });
+  } catch (err) {
```
aiwilliams, line 66: I think we'll want a `findEntity(key)` that doesn't throw an error as another option someday.

aiwilliams:
Only thing missing is CHANGELOG, but we can get that on the release.

PR #119, opened 2020-08-03
Add service principals

ndowmon:


src/steps/active-directory/converters.ts
```diff
@@ -112,6 +115,30 @@ export function createUserEntity(data: User): Entity {
   });
 }
 
+export function createServicePrincipalEntity(data: any): Entity {
+  const entity = createIntegrationEntity({
+    entityData: {
+      source: data,
+      assign: {
+        _key: generateEntityKey(SERVICE_PRINCIPAL_ENTITY_TYPE, data.id),
```
aiwilliams, line 123: I was thinking we'd avoid including the `_type` value in the `_key` when possible, but I might be forgetting something 😬

ndowmon: Yes, we are. I should have commented about this.

The direct relationship mapper for `roleAssignment`s had been set up for the `User` and `Group` entities, both of which use `generateEntityKey()`. I was trying to avoid making any more conditional logic on the `roleAssignment` code so I kept the same pattern here. 

Soon I can go through and tease out all refs to `generateEntityKey` if it's safe to do so.

PR #122, opened 2020-08-05
Remove _type prefix on generateEntityKey

ndowmon:


src/utils/generateKeys.ts
```diff
@@ -5,7 +5,7 @@ export function generateEntityKey(
   if (!id) {
     throw new Error('Cannot generate entity key with `undefined` id');
   }
-  return `${type}_${id}`;
```
austinkelleher, line 8: Can we remove `type` as an arg to this function now?

aiwilliams: Ah, yes, this should be taken care of before merging.

ndowmon: Resolved

ndowmon: Changed the interface to remove `type`.

src/utils/generateKeys.ts
```diff
@@ -1,11 +1,8 @@
-export function generateEntityKey(
-  type: string,
-  id: string | number | undefined,
-): string {
+export function generateEntityKey(id: string | number | undefined): string {
```
aiwilliams, line 1: I think eventually we should not need this because `createIntegrationEntity` should ensure we end up with a valid `_key` value.

CHANGELOG.md
```diff
@@ -22,7 +26,7 @@ and this project adheres to
 - Added `azure_role_definition|assigned|<type>` relationships
 - Added `azure_classic_admin_group` singleton entity
 - Added `azure_classic_admin_group|has|azure_user` relationships
-- Added `azure_service_principal` entities
+- Added `azuer_service_principal` entities
```
austinkelleher, line 29: Typo

Issue #127, opened 2020-08-20
Question-driven development: Azure Permissions


ndowmon:
I'm trying to figure out the best way to add entities/relationships so that we can answer useful questions for azure users.

Given the following goal: 
**Find all azure principals that have the ability to delete key vaults in the `ndowmon-j1dev` key vault**

Pseudocode:
```
FIND principal (user, user group, or service principal) \\ could be any of the three
  THAT IS ASSIGNED roleDefinition
  THAT (ALLOWS?) scope (management group, subscription, resource group, or resource) 
  WHERE roleDefinition.notAllows !== Microsoft.KeyVault/vaults/delete and 
        roleDefinition.allows === Microsoft.KeyVault/vaults/delete and
        resource.name === ndowmon-j1dev
```

```
FIND (azure_user|azure_user_group|azure_service_principal)
   THAT ASSIGNED azure_role_definition
   THAT ALLOWS (
      azure_keyvault WITH name = key-vault-id OR
      azure_resource_group THAT HAS azure_keyvault WITH name = key-vault-id OR
      azure_subscription THAT HAS azure_resource_group THAT HAS azure_keyvault WITH name = key-vault-id
   )
   WHERE 
      azure_role_definition.notAllows !== (
         * OR 
         */delete OR 
         Microsoft.KeyVault/*/delete OR 
         Microsoft.KeyVault/vaults/delete
      ) AND 
      azure_role_definition.allows === (
         * OR
         */delete OR
         Microsoft.KeyVault/*/delete OR 
         Microsoft.KeyVault/vaults/delete
      )
```

Questions:
1. Should the services on azure_role_definition.allows become an entity? I think they should just be fields on azure_role_definition but maybe it would be better to say something like `THAT ALLOWS Microsoft.KeyVault WITH notAllows !== (* OR */delete OR vaults/delete) AND allows == (* OR */delete OR vaults/delete`.
1. Does the J1QL allow for a more elegant parsing syntax so that we wouldn't need to use `(* OR */delete ...)` for string matching these RBAC permissions?
1. Does J1QL implement any traversal shortcuts for the target scope (keyvault, resource group or subscription)? Something like the [* (0 or more)](https://chortle.ccsu.edu/FiniteAutomata/Section07/sect07_16.html) quantifier in regular expressions, e.g. `THAT ALLOWS (* THAT HAS)* azure_keyvault WITH name = key-vault-id`.
[...]

PR #128, opened 2020-08-21
J1 4385 azure role assignments

ndowmon:
Related to #127, update the azure iam resources to create `azure_role_assignment` permissions, as opposed to making role assignments a node on the graph.


docs/jupiterone.md
```diff
@@ -93,7 +93,8 @@ though the `Account` entity will always be ingested.
 |                      | `azure_sql_server`            | `Database`, `DataStore`, `Host` |
 | Cosmos DB            | `azure_cosmosdb_account`      | `Account`                       |
 |                      | `azure_cosmosdb_sql_database` | `Database`, `DataStore`         |
-| Role Definition      | `azure_role_definition`       | `AccessRole`, `AccessPolicy`    |
+| Role Definition      | `azure_role_definition`       | `AccessRole`                    |
```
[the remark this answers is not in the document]
ndowmon: Thanks for the catch! You're right, here's a classic example of docs not matching implementation.

I'll open a follow-on PR for version 3.0.0

terraform/iam.tf
```diff
@@ -15,3 +15,17 @@ resource "azurerm_role_assignment" "j1dev" {
   role_definition_id = azurerm_role_definition.j1dev_subscription.id
   principal_id       = data.azurerm_client_config.current.object_id
 }
+
+resource "azurerm_role_assignment" "j1dev-resource-group" {
+  name               = "20000000-0000-0000-0000-000000000000"
```
austinkelleher, line 20: What was the motivation of naming like this?

ndowmon: @austinkelleher the name must be UUID format and gets rejected if not hex characters in sequence <8>-<4>-<4>-<4>-<12>. This field is basically the only identifier I can reliably use in client tests, so I wanted to use a sequence that's visually simple to validate.

PR #138, opened 2020-08-31
Add azure subscriptions

ndowmon:
Adds the following:
- `azure_subscription` entities (should be one, considering we authenticate with a `subscriptionId`
- `azure_subscription--HAS-->azure_resource_group` relationships

Also updates the `role_assignment--ALLOWS-->azure_subscription` relationship, which was previously mapped.

Find a graph of `azure_subscription--HAS-->azure_resource_group--HAS--><resource>` below: 
Screen Shot 2020-08-31 at 2 02 36 PM

### DIFF
--- a/src/steps/resource-manager/authorization/constants.ts
+++ b/src/steps/resource-manager/authorization/constants.ts
@@ -28,6 +28,10 @@ import {
   EOL_MATCHER,
   RESOURCE_GROUP_MATCHER,
 } from '../utils/matchers';
+import {
+  SUBSCRIPTION_ENTITY_METADATA,
+  STEP_RM_SUBSCRIPTIONS,
+} from '../subscriptions';
 
 // Fetch Role Assignments
 export const STEP_RM_AUTHORIZATION_ROLE_ASSIGNMENTS =
@@ -154,8 +158,8 @@ export const SCOPE_TYPES_MAP: ResourceIdMap[] = [
   ...RESOURCE_ID_TYPES_MAP,
   {
     resourceIdMatcher: new RegExp(SUBSCRIPTION_MATCHER + EOL_MATCHER),
-    _type: 'azure_subscription',
-    dependsOn: [],
+    _type: SUBSCRIPTION_ENTITY_METADATA._type,
+    dependsOn: [STEP_RM_SUBSCRIPTIONS],
   },
   {
     resourceIdMatcher: new RegExp(RESOURCE_GROUP_MATCHER + EOL_MATCHER),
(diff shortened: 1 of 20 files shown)


src/steps/resource-manager/subscriptions/client.ts
```diff
@@ -0,0 +1,35 @@
+import { SubscriptionClient } from '@azure/arm-subscriptions';
+import { Subscription } from '@azure/arm-subscriptions/esm/models';
+import { Client } from '../../../azure/resource-manager/client';
+import { IntegrationProviderAPIError } from '@jupiterone/integration-sdk-core';
+
+export class J1SubscriptionClient extends Client {
+  public async iterateSubscriptions(
+    callback: (s: Subscription) => void | Promise<void>,
+  ): Promise<void> {
+    const serviceClient = await this.getAuthenticatedServiceClient(
```
ndowmon, line 10: Unlike other clients, which take `{ auth, subscriptionId, options }` args, the subscription client takes only `{ auth, options }`. Hence these changes to `getAuthenticatedServiceClient`.

src/steps/resource-manager/authorization/constants.ts
```diff
@@ -154,8 +158,8 @@ export const SCOPE_TYPES_MAP: ResourceIdMap[] = [
   ...RESOURCE_ID_TYPES_MAP,
   {
     resourceIdMatcher: new RegExp(SUBSCRIPTION_MATCHER + EOL_MATCHER),
-    _type: 'azure_subscription',
-    dependsOn: [],
+    _type: SUBSCRIPTION_ENTITY_METADATA._type,
+    dependsOn: [STEP_RM_SUBSCRIPTIONS],
```
aiwilliams, line 162: I find myself wondering if this should be a singular form, since the integration will not ingest multiple subscriptions. 🤔

ndowmon: Eh, we are technically calling a `list` API, so I think we should still leave it plural. In case more are for some reason returned.

This is chained on https://github.com/JupiterOne/graph-azure/pull/137 so I'll wait to merge.

src/steps/resource-manager/subscriptions/client.ts
```diff
@@ -0,0 +1,35 @@
...
+      SubscriptionClient,
+      {
+        passSubscriptionId: false,
+      },
+    );
+    try {
+      const items = await serviceClient.subscriptions.list();
```
austinkelleher, line 17: I'm guessing this always returns all data and does not need to be paginated?

ndowmon: Actually I should be using the `iterateAllResources` method that Adam implemented. making the change

PR #149, opened 2020-09-02
Fixes #146 - Stop creating mapped role_assignment->scope relationships

### LINKED ISSUE #146: Match roleAssignment -> scopes _only_ where explicit
I have discussed with @aiwilliams that we don't want to generate  any placeholder entities for role assignment -> scope targets. If the entity is not found, we should not create a relationship.

The simple implementation creates a problem: what happens if we have a role assignment that points to a _known_ entity type (ex. `azure_keyvault_service`) that for some reason is not found with `jobState.findEntity(keyvaultId)`? We want to log that the entity was not found, so developers can see if there is something wrong with the implementation of `role_assignment->keyvault_service` relationships.

Simply logging any time an entity is _not_ found, however, will create noise for many, many entities that we're not ingesting yet. So we should first check if the resourceId matches a pattern we expect to ingest before looking for that entity. (`/subscriptions/<s-id>/resourceGroups/<rg-id>/providers/Microsoft.KeyVault/vaults/<v-id>`)

However, if we only create relationships when we _explicitly_ know the matching resource ID pattern, it requires that developers always add matching patterns for new resource entities. If an entity does exist in the job state, but there is not matcher, we should alert developers so they can add the new matching pattern and create relationships between `role_assignment` and the new entity type.

Since we want to be able to check if a target `scope` exists in the `jobState`, the step which generates `role-assignment->scope` relationships needs to run as late as possible in the dependency tree. This would require a change to the SDK.

ndowmon:


CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Removed
+
+- Removed mapped `role_assignment|allows|<scope>` relationships
```
aiwilliams, line 13: Do you think it'd be worth pointing this out:
```suggestion
- Removed mapped `role_assignment|allows|<scope>` relationships to avoid creating `azure_unknown_resource_type` entities
```

ndowmon: Thanks for the suggestion, like it!

PR #150, opened 2020-09-02
Fix storage account relationships

ndowmon:
Storage accounts have direct relationships to **Blobs**, **File Shares**, **Queues**, **Tables** and in some cases **Data Lakes** or other storage services. 

A storage account has a **Kind** property which can limit the types of storage services that can exist within a storage account,  including **Storage**, **BlobStorage,** **StorageV2**, and others. 

Previously, a storage account that had some **Blobs** and some **File Shares** would create two entities to represent the storage account. This PR creates a single `azure_storage_account` entity per storage account, and links the storage account to the `azure_resource_group` associated with the storage account.

Old graph:
Screen Shot 2020-09-02 at 5 10 23 PM

New Graph:
Screen Shot 2020-09-02 at 5 08 47 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,10 +10,13 @@ and this project adheres to
 
 ### Removed
 
-- Removed mapped `role_assignment|allows|<scope>` relationships to avoid creating `azure_unknown_resource_type` entities
+- Removed mapped `role_assignment|allows|<scope>` relationships to avoid
+  creating `azure_unknown_resource_type` entities
 
 ### Changed
 
+- Created `azure_storage_account` entities to replace
+  `azure_storage_blob_service` and `azure_storage_file_service`
 - Upgraded SDK to v3.2.0, ordered entity/relationship docs
 
 ## 4.4.1 - 2020-09-02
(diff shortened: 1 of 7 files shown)


aiwilliams (approved):

aiwilliams:
@ndowmon It may be worth noting in the CHANGELOG, something like: Any queries that depend on `Service` entities must be changed.

ndowmon:
FYI this is a breaking change since relationships are being re-written. It _may_ cause customer impact for any customers who have written custom Azure queries that include traversals between the impacted entities. This should be noted in the next set of release notes. cc @erkangz

aiwilliams:
Also, could be nice to get those images into the release notes, though I'm not sure it wouldn't confuse the casual observer. Something to think about as we look to improve how we communicate the impact of changes.

aiwilliams:
> Any queries that depend on `Service` entities must be changed.

@ndowmon I was wrong about this, I see now that the `_type` was the change, the `_class` is still `Service`. However, it may still be helpful to consider what statement should be made, if any, about how any queries may need updating.

PR #151, opened 2020-09-03
Add storage queues

ndowmon:
Added `azure_storage_queue`
Added `azure_storage_account--HAS-->azure_storage_queue`


src/steps/resource-manager/storage/client.ts
```diff
@@ -5,10 +5,67 @@ import {
...
+  resourceGroupName,
+  accountName,
+}: IterateAllStorageAccountResourcesOptions<ResourceType>) {
+  return iterateAllResources({
+    logger,
+    serviceClient,
+    resourceEndpoint: {
+      list: async () => {
+        try {
+          const resources = await resourceEndpoint.list(
```
ndowmon, line 38: `StorageManagementClient.queue`, `.table`, `.container`, and `.fileShare` all require `resourceGroupName` and `storageAccountName` as args to the `.list()` method. 

I added this intermediate `iterateAllStorageAccountResources` that can be leveraged by all children/resources of `StorageAccount`

src/steps/resource-manager/storage/client.ts
```diff
@@ -5,10 +5,67 @@ import {
...
+            resourceGroupName,
+            accountName,
+          );
+          return resources;
+        } catch (err) {
+          if (err?.body?.code === 'FeatureNotSupportedForAccount') {
```
ndowmon, line 44: Since we iterate all storage accounts, but some storage accounts cannot contain certain types of data stores (e.g. storage account with `kind=BlobStorage` cannot contain `queues`), allow execution to continue and return no queues when `FeatureNotSupportedForAccount`

src/steps/resource-manager/storage/client.ts
```diff
@@ -5,10 +5,67 @@ import {
...
+            response._response = { request: err?.request };
+            return response;
+          }
+          throw err;
+        }
+      },
+      listNext: /* istanbul ignore next: testing iteration might be difficult */ async (
+        nextLink: string,
+      ) => {
+        return serviceClient.blobContainers.listNext(nextLink);
```
aiwilliams, line 56: I'm not 100% sure, but this line looks problematic because the containing function seems independent of the kind, but this is referencing `blobContainers`.

src/steps/resource-manager/storage/client.ts
```diff
@@ -57,6 +114,28 @@ export class StorageClient extends Client {
     });
   }
 
+  /* eslint-disable @typescript-eslint/no-non-null-assertion */
+  public async iterateQueues(
+    storageAccount: { name: string; id: string },
+    callback: (e: BlobContainer) => void | Promise<void>,
```
aiwilliams, line 120: `BlobContainer`?

src/steps/resource-manager/storage/converters.ts
```diff
@@ -124,3 +126,32 @@ export function createStorageFileShareEntity(
     },
   });
 }
+
+/**
+ * Creates an integration entity for a Files service file share.
+ *
+ * * Files are considered to be encrypted when the storage account is
+ *   configured as encrypted. See
+ *   https://azure.microsoft.com/en-us/blog/announcing-default-encryption-for-azure-blobs-files-table-and-queue-storage/
```
aiwilliams, line 135: Copypasta?

aiwilliams (changes requested):

ndowmon:
@aiwilliams thanks for the review. Made requested changes, can you take another look?

src/steps/resource-manager/storage/client.ts
austinkelleher: Is there a way that we can make this a bit more explicit so that we do not perform a bunch of API calls that we know will fail?

austinkelleher (approved):

aiwilliams (approved):
Looks great. So sorry to have missed that that Azure recording function was not used before! I had wondered at one point if there were a way to avoid that mistake. Seems like we might need to learn more about creating eslint rules. 😄

PR #152, opened 2020-09-03
Add storage tables

ndowmon:
Added `azure_storage_table`
Added `azure_storage_account--HAS-->azure_storage_table`


src/steps/resource-manager/storage/client.ts
```diff
@@ -41,7 +42,12 @@ async function iterateAllStorageAccountResources<ResourceType>({
           );
           return resources;
         } catch (err) {
-          if (err?.body?.code === 'FeatureNotSupportedForAccount') {
+          if (
+            [
+              'OperationNotAllowedOnKind',
```
ndowmon, line 47: I encountered a different error code that's thrown when a `BlobStorage` account tries to iterate over `Table` records.

src/steps/resource-manager/storage/constants.ts
```diff
@@ -72,3 +72,22 @@ export const STORAGE_ACCOUNT_QUEUE_RELATIONSHIP_METADATA = {
   _class: STORAGE_ACCOUNT_RELATIONSHIP_CLASS,
   targetType: STORAGE_QUEUE_ENTITY_METADATA._type,
 };
+
+// Storage Tables
+export const STEP_RM_STORAGE_TABLES = 'rm-storage-tables';
+export const STORAGE_TABLE_ENTITY_METADATA = {
+  _type: 'azure_storage_table',
+  _class: ['DataStore'],
```
ndowmon, line 80: Not sure if `DataStore` is a good entity type here. Seemed more generic to me than `Database` but others may disagree. Open to suggestions.

Maybe there should be a `Table` entity class.

src/steps/resource-manager/storage/client.test.ts
```diff
@@ -238,3 +238,44 @@ describe('iterateQueues', () => {
...
+
+  test('all', async () => {
+    recording = setupRecording({
+      directory: __dirname,
+      name: 'iterateTables',
+    });
+
+    const client = new StorageClient(config, createMockIntegrationLogger());
+
+    const resources: StorageQueue[] = [];
```
aiwilliams, line 259: `StorageQueue`?

src/steps/resource-manager/storage/client.ts
aiwilliams: Hmm, wondering if this is because the function references `blobContainers`? See https://github.com/JupiterOne/graph-azure/pull/151#discussion_r483160140.

src/steps/resource-manager/storage/constants.ts
aiwilliams: I'd copy whatever AWS is doing for DynamoDB tables.

aiwilliams (changes requested):

src/steps/resource-manager/storage/constants.ts
ndowmon: AWS uses `[ 'Database', 'DataStore' ]` - good tip.

src/steps/resource-manager/storage/client.test.ts
ndowmon: Good catch, thanks. Almost identical interfaces but still.

src/steps/resource-manager/storage/client.ts
ndowmon: This error is actually not called in the same branch that used `blobContainers`, but I did check after fixing and it still exists.

My guess is that whoever wrote the api for `StorageQueues` was different or just didn't share error code with whoever wrote it for `Tables`. Not sure. But I'm going to leave this as is since the error code is definitely showing up in my API calls.

austinkelleher (dismissed):

aiwilliams (dismissed):

austinkelleher (approved):

PR #156, opened 2020-09-08
Add api management

ndowmon:
- Added `azure_api_management_service` entities
- Added `azure_resource_group|has|azure_api_management_service` relationships
- Added `azure_api_management_api` entities
- Added `azure_api_management_service|has|azure_api_management_api`
  relationships

**EXAMPLE**
Screen Shot 2020-09-08 at 4 22 30 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,14 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Added
+
+- Added `azure_api_management_service` entities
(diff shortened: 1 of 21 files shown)


ndowmon:
I felt that the `Gateway` class fit for `azure_api_management_service` and the `ApplicationEndpoint` class fit for `azure_api_management_api`. Interested to hear what you all think about those classes.

aiwilliams (approved):

PR #158, opened 2020-09-09
Add dns

ndowmon:
Screen Shot 2020-09-09 at 5 53 09 PM

### DIFF
--- a/src/steps/resource-manager/dns/client.ts
+++ b/src/steps/resource-manager/dns/client.ts
@@ -1,0 +1,57 @@
+import { DnsManagementClient } from '@azure/arm-dns';
+import { Zone, RecordSet } from '@azure/arm-dns/esm/models';
+import {
+  Client,
+  iterateAllResources,
+  ListResourcesEndpoint,
+} from '../../../azure/resource-manager/client';
+import { resourceGroupName } from '../../../azure/utils';
+
+export class J1DnsManagementClient extends Client {
+  public async iterateDnsZones(
+    callback: (s: Zone) => void | Promise<void>,
+  ): Promise<void> {
+    const serviceClient = await this.getAuthenticatedServiceClient(
+      DnsManagementClient,
+    );
+
+    return iterateAllResources({
+      logger: this.logger,
+      serviceClient,
+      resourceEndpoint: serviceClient.zones,
+      resourceDescription: 'dns.zone',
+      callback,
+    });
+  }
+
+  public async iterateDnsRecordSets(
+    dnsZone: { name: string; id: string },
+    callback: (s: RecordSet) => void | Promise<void>,
+  ): Promise<void> {
+    const serviceClient = await this.getAuthenticatedServiceClient(
+      DnsManagementClient,
+    );
+    const resourceGroup = resourceGroupName(dnsZone.id, true)!;
+    const zoneName = dnsZone.name;
+
+    return iterateAllResources({
+      logger: this.logger,
+      serviceClient,
+      resourceEndpoint: {
+        list: async () => {
+          return serviceClient.recordSets.listByDnsZone(
+            resourceGroup,
+            zoneName,
+          );
+        },
+        listNext: /* istanbul ignore next: testing iteration might be difficult */ async (
+          nextLink: string,
+        ) => {
+          return serviceClient.recordSets.listByDnsZoneNext(nextLink);
+        },
+      } as ListResourcesEndpoint,
+      resourceDescription: 'dns.recordSet',
+      callback,
+    });
+  }
+}
--- a/src/steps/resource-manager/dns/constants.ts
+++ b/src/steps/resource-manager/dns/constants.ts
@@ -1,0 +1,38 @@
+import {
+  generateRelationshipType,
+  RelationshipClass,
+} from '@jupiterone/integration-sdk-core';
+import { createResourceGroupResourceRelationshipMetadata } from '../utils/createResourceGroupResourceRelationship';
+
+// Fetch Resource Groups
+export const STEP_RM_DNS_ZONES = 'rm-dns-zones';
+export const STEP_RM_DNS_RECORD_SETS = 'rm-dns-record-sets';
+
+export const DnsEntities = {
+  ZONE: {
+    _type: 'azure_dns_zone',
+    _class: ['DomainZone'],
+    resourceName: '[RM] DNS Zone',
+  },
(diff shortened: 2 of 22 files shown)


src/steps/resource-manager/dns/constants.ts
```diff
@@ -0,0 +1,38 @@
+import {
+  generateRelationshipType,
+  RelationshipClass,
+} from '@jupiterone/integration-sdk-core';
+import { createResourceGroupResourceRelationshipMetadata } from '../utils/createResourceGroupResourceRelationship';
+
+// Fetch Resource Groups
```
aiwilliams, line 7: I'm not following the meaning of this comment.

src/steps/resource-manager/dns/converters.test.ts
```diff
@@ -0,0 +1,65 @@
...
+        'ns3-05.azure-dns.org.',
+        'ns4-05.azure-dns.info.',
+      ],
+      zoneType: 'Public',
+    };
+
+    expect(createDnsZoneEntity(webLinker, data)).toMatchSnapshot();
+    expect(createDnsZoneEntity(webLinker, data)).toMatchGraphObjectSchema({
+      _class: ['DomainZone'],
+      schema: {},
```
aiwilliams, line 31: Did you mean to add something to the schema?

src/steps/resource-manager/dns/converters.test.ts
```diff
@@ -0,0 +1,65 @@
...
+        retryTime: 300,
+        expireTime: 2419200,
+        minimumTtl: 300,
+      },
+    };
+
+    expect(createDnsRecordSetEntity(webLinker, data)).toMatchSnapshot();
+    expect(createDnsRecordSetEntity(webLinker, data)).toMatchGraphObjectSchema({
+      _class: ['DomainRecord'],
+      schema: {},
```
aiwilliams, line 62: Another empty schema.

src/steps/resource-manager/dns/client.ts
```diff
@@ -0,0 +1,57 @@
+import { DnsManagementClient } from '@azure/arm-dns';
+import { Zone, RecordSet } from '@azure/arm-dns/esm/models';
+import {
+  Client,
+  iterateAllResources,
+  ListResourcesEndpoint,
+} from '../../../azure/resource-manager/client';
+import { resourceGroupName } from '../../../azure/utils';
+
+export class J1DnsManagementClient extends Client {
```
aiwilliams, line 10: I'm curious, why name this `J1DnsManagementClient` instead of `DnsClient`? Consider the [`ComputeClient`](https://github.com/JupiterOne/graph-azure/blob/47361544db03cfc7f6fb7bb3a359c5c19fb51eaa/src/steps/resource-manager/compute/client.ts#L14) for example. The reason I ask is because the `J1` prefix sent me on a path to discover how it was special.

aiwilliams (commented):
💥 Adding things so quick now that the patterns are established! Thanks a ton.

src/steps/resource-manager/dns/constants.ts
ndowmon: Oops. typo!

src/steps/resource-manager/dns/converters.test.ts
ndowmon: In this iteration I switched to snapshot testing, and I _also_ haven't explicitly set any properties except those required by the `DomainZone` schema.

```
export function createDnsZoneEntity(
  webLinker: AzureWebLinker,
  data: Zone,
): Entity {
    entityData: {
      source: data,
      assign: {
        ...convertProperties(data),
        _key: data.id as string,
        _type: DnsEntities.ZONE._type,
        _class: DnsEntities.ZONE._class,
        id: data.id,
        name: data.name,
        domainName: data.name,
        recordsCount: data.numberOfRecordSets,
      },
    },
  });
```

I don't really think that schemas should be set if we're not explicitly requiring certain properties to be added, and the `DomainZone` properties (`domainName`, `recordsCount`, `name`, `id`) are already validated by the jest matcher. 

It's a new philosophy for testing but I'd like to know your thoughts. I think it's one that will increase our velocity and reduce friction, since we aren't _requiring_ any other properties in our schema.

src/steps/resource-manager/dns/client.ts
ndowmon: That was a pattern I introduced in https://github.com/JupiterOne/graph-azure/pull/156. There, the _service_ name is **API Management**, and `@azure/arm-apimanagement` client is called `ApiManagementClient`. Following our current pattern, which would name our internal client `ApiClient` didn't make sense in that instance, so I instead prefixed with `J1`. 

I guess the idea is that prefixing the client with `J1` rather than removing `Management` can be done consistently without losing meaning in the client names. But I recognize that the pattern change makes things inconsistent with what's already there.

src/steps/resource-manager/dns/constants.ts
ndowmon: This is fixed.

src/steps/resource-manager/dns/converters.test.ts
austinkelleher: This makes sense to me. I see why it's confusing, but fundamentally makes sense because the underlying parent schemas are all being merged. Really what we care about is the child properties that this specific entity `_type` has defined. In this case, there are no new properties.

Perhaps we should make the interface more clear that the schema is the child's schema?

austinkelleher (approved):

src/steps/resource-manager/dns/converters.test.ts
aiwilliams: I think we should change `toMatchGraphObjectSchema` to make the `schema` property optional. I would not have though twice about it if an empty value had not been passed. If that makes sense, would you mind creating an Issue in SDK @ndowmon ? Thanks!

src/steps/resource-manager/dns/client.ts
aiwilliams: Makes sense, thanks for the explanation. Perhaps we could rename the other clients to match this better pattern in a future PR.

aiwilliams (approved):

src/steps/resource-manager/dns/client.ts
ndowmon: Yep. I'd like to go back and make this change and some changes to the `constants.ts` files a little later this week for consistency.

PR #159, opened 2020-09-09
Add private dns

ndowmon:
Virtually identical to #158 

Screen Shot 2020-09-09 at 7 19 23 PM


aiwilliams (commented):
Same question re: empty `schema`s in tests.

It's surprising on one hand that private DNS is a completely different service, but on the other hand there is beauty in consistency.

aiwilliams (approved):

PR #162, opened 2020-09-20
Make subscription id optional

ndowmon:
This change allows users to set `ingestResourceManager` to `false`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to
 
 ### Added
 
+- Added `ingestResourceManager` flag
 - Added `azure_dns_zone` entities
 - Added `azure_resource_group|has|azure_dns_zone` relationships
 - Added `azure_dns_record_set` entities
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to
 
 ## [Unreleased]
 
+## 5.2.0 - 2020-09-20
+
 ### Added
 
 - Added `ingestResourceManager` flag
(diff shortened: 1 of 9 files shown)


src/getStepStartStates.ts
```diff
@@ -115,4 +113,18 @@ export default function getStepStartStates(
     [STEP_RM_CONTAINER_REGISTRIES]: resourceManager,
     [STEP_RM_CONTAINER_REGISTRY_WEBHOOKS]: resourceManager,
   };
+
+  executionContext.logger.info(
```
austinkelleher, line 117: Did you intend for this to be exposed to the customer? You have to use `logger.publishEvent` to expose to customer integration log.

austinkelleher (approved):
Besides minor question about logging, this looks good to me. Thanks for fixing!

CHANGELOG.md
```diff
@@ -8,8 +8,11 @@ and this project adheres to
 
 ## [Unreleased]
 
+## 5.2.0 - 2020-09-20
+
 ### Added
 
+- Added `ingestResourceManager` flag
```
austinkelleher, line 15: I think it would beneficial to explain what this is used for in the CHANGELOG

src/getStepStartStates.ts
ndowmon: Yeah, that was a bit of an oversight. I agree with what you said in https://github.com/JupiterOne/graph-fastly/pull/3#pullrequestreview-492570290 so I can remove this log statement in a future iteration.

CHANGELOG.md
ndowmon: In the spirit of getting this change out, I'll explain this flag in the upcoming https://github.com/JupiterOne/graph-azure/pull/161.

aiwilliams (commented):
I'd like to get https://github.com/JupiterOne/graph-microsoft-365/tree/initial-release moving forward so that folks can install a Microsoft 365 integration. I'll add an Epic for this so we can prioritize alongside other work.

PR #166, opened 2020-09-24
Feature/azure event grid

KeionneDerousselle:
# Acceptance Criteria
- Map the entities and relationships from Azure Event Grid for an Azure Subscription.

# What Changed?

1. Added the **Event Grid Domain** entity
A Domain is a custom resource that can be added to a Resource Group that represents a domain or area of concern for a particular service or organization.

2. Added the **Event Grid Domain Topic** entity
A Domain Topic is an event queue that holds events relating to a Domain.

3. Added the **Event Grid Topic** Entity
A Topic is an event queue that holds events relating to a specific topic. Topics are not necessarily related to a domain and can simply belong to a Resource Group.

4. Added the **Event Grid Topic Subscription** entity
A Topic Subscription is a subscription to receive events from a Topic queue. Topic Subscriptions can be a subscription to any Topic, not necessarily a Domain Topic. In fact, one can subscribe to just about any resource in Azure as a Topic. https://docs.microsoft.com/en-us/cli/azure/eventgrid/event-subscription?view=azure-cli-latest

    You can subscribe to events about Azure Storage Accounts, other Azure Subscriptions, Azure Resource Groups, or any other Azure resource (e.g. Microsoft.ContainerRegistry, Microsoft.Batch). 

    Event Grid Topic Subscriptions to custom Event Grid Topics, generic Event Grid Topics, or other regional resources such as an Azure Storage Account are known as **regional** subscriptions. Event Grid Topic Subscriptions to global resources such as Azure Subscriptions or Azure Resource Groups are known as **global** subscriptions.

    For the purposes of this PR, only the relationship between Event Grid Topic Subscriptions to custom or generic Event Grid Topics are mapped. Because the relationship between Event Grid Topic Subscriptions to other Azure resources is much more complex, it will be addressed in another PR.

5. Added the **Azure Resource Group** has **Event Grid Domain** relationship

6. Added the **Event Grid Domain** has **Event Grid Domain Topic** relationship

7. Added the **Event Grid Domain Topic** has **Event Grid Topic Subscription** relationship

8. Added the **Azure Resource Group** has **Event Grid Topic** relationship

9. Added the **Event Grid Topic** has **Event Grid Topic Subscription** relationship

# Graph
event grid relationships and entities


ndowmon:
@KeionneDerousselle FYI, this is not well documented, and it looks like you got it figured out, but you should run the following scripts locally (sorry I didn't mention it sooner!):

```bash
### runs linting via ESLint
yarn lint 

### runs typescript compile (`tsc`)
yarn type-check

### runs jest
yarn test
```

Our github workflows run these same scripts for open PRs and requires these checks to pass before a branch is allowed to be merged. So I usually run `yarn lint && yarn type-check && yarn test` in the command line before pushing changes. Tribal knowledge...

KeionneDerousselle:
@ndowmon Thanks for the heads up! I could also probably set up linting on editor save.

src/steps/resource-manager/event-grid/converters.test.ts
```diff
@@ -0,0 +1,155 @@
...
+      },
+      type: 'Microsoft.EventGrid/domains',
+    };
+
+    const domainEntity = createEventGridDomainEntity(webLinker, data);
+
+    expect(domainEntity).toMatchSnapshot();
+    expect(domainEntity).toMatchGraphObjectSchema({
+      _class: ['Service'],
+      schema: {},
```
aiwilliams, line 41: I think the latest versions of the SDK testing package have `schema` as an optional value.

src/steps/resource-manager/event-grid/index.ts
```diff
@@ -0,0 +1,296 @@
...
+        await client.iterateDomainTopicSubscriptions(
+          ({
+            resourceGroupName: resourceGroup,
+            name,
+            domainName,
+          } as unknown) as {
+            resourceGroupName: string;
+            name: string;
+            domainName: string;
+          },
```
aiwilliams, line 136: This isn't blocking at all, but....

I wonder if there's a way to avoid the cast to `unknown` 🤔 I see this is a pattern in other fetching functions. I wonder why TypeScript cannot infer the type of the arg in `iterateDomainTopicSubscriptions(param0)`.

Related, I see something similar here:

https://github.com/JupiterOne/graph-azure/blob/eb9bf0e9d16746173051d93f473297575f82fc8c/src/steps/resource-manager/storage/index.ts#L212

@ndowmon I think we should consider how `iterateEntities` could be improved to support something like:
```ts
await jobState.iterateEntities<{ ... }>({ ... }, async (entity) => {
  // entity will now be typed as the generic type provided to iterateEntities<T>
});
```
where `T` is `Entity & { ... }`.

terraform/iam.tf
```diff
@@ -12,55 +12,55 @@ resource "azurerm_role_definition" "j1dev_subscription" {
 resource "azurerm_role_assignment" "j1dev" {
   name               = "10000000-0000-0000-0000-000000000000"
   scope              = data.azurerm_subscription.j1dev.id
-  role_definition_id = azurerm_role_definition.j1dev_subscription.id
+  role_definition_id = split("|", azurerm_role_definition.j1dev_subscription.id)[0]
```
aiwilliams, line 15: I'm surprised that the output of the `id` property from the `azurerm_role_definition` resource is not longer directly usable in the `azurerm_role_assignment` resource. Perhaps something has changed in a later version of the Azure Terraform provider so that this is now necessary. @ndowmon @KeionneDerousselle, could you compare the versions of the provider that you both have?

We should add version constraints here (a best practice anyhow, can be a different PR):

https://github.com/JupiterOne/graph-azure/blob/eb9bf0e9d16746173051d93f473297575f82fc8c/terraform/main.tf#L1

There is a `.terraform-version` file that affects `tfenv`, but it doesn't control the version of the providers.

src/steps/resource-manager/event-grid/constants.ts
```diff
@@ -0,0 +1,84 @@
...
+      EventGridEntities.TOPIC_SUBSCRIPTION,
+    ),
+
+    sourceType: EventGridEntities.TOPIC._type,
+    _class: RelationshipClass.HAS,
+    targetType: EventGridEntities.TOPIC_SUBSCRIPTION._type,
+  },
+
+  // TODO : Figure out what kinds of subscriptions we want to capture.
+  // global, regional, subscription (account level), resource group
```
aiwilliams, line 83: I saw this in the Issue opening comment:

> For the purposes of this PR, only the relationship between Event Grid Topic Subscriptions to custom or generic Event Grid Topics are mapped. Because the relationship between Event Grid Topic Subscriptions to other Azure resources is much more complex, it will be addressed in another PR.

That is an excellent decision. Do you think it wise to talk about this more in a call?

aiwilliams (approved):
This looks great @KeionneDerousselle! I'm going to approve because I don't think any comments are blocking, though would appreciate follow up on them so we can get other Issues created as necessary.

src/steps/resource-manager/event-grid/constants.ts
KeionneDerousselle: Sure thing!

terraform/iam.tf
KeionneDerousselle: @aiwilliams This is a bug introduced the version of terraform I installed. https://github.com/terraform-providers/terraform-provider-azurerm/issues/8547

KeionneDerousselle: @ndowmon Did not experience this issue. I can remove this change from the PR or lockdown version constraints on the Azure Terraform provider. (I might need help with the latter. I'm not very familiar with terraform)

src/steps/resource-manager/event-grid/index.ts
KeionneDerousselle: I was trying to avoid passing the entire entity to the client call. I was casting it to a generic object so I only received the properties I needed in the client calls. Is there a better way to do this? Sorry, I'm new to typescript.

src/steps/resource-manager/event-grid/converters.test.ts
KeionneDerousselle: So can I just remove the schema property from these assertions?

terraform/iam.tf
KeionneDerousselle: It looks that bug has been resolved. I will try updating :)

KeionneDerousselle: I tried locking the version to the latest version, but it still had the '|' issue.

src/steps/resource-manager/event-grid/converters.test.ts
KeionneDerousselle: I tried upgrading the integrations sdk to 3.3.0, 

updates

but when running the tests, I got some errors. I wasn't sure whether I should try to map the existing types to the new ones or if that was a risky move.

errors

src/steps/resource-manager/event-grid/index.ts
ndowmon: It's totally fair to pass the entire object to the client call, I would say it's preferred in that it is succinct.

However, that wouldn't be strictly possible in this case, since only the `name` property comes from the entity you're passing in to iterate, and you're doing some manual logic to get `resourceGroupName` and `domainName` before passing them to the client.

The pattern we usually follow is to do the logic of getting `domainName` and `resourceGroupName` within the client itself, so that your only argument to the client call is an existing entity. See the following from the `cdn` resoures: 


```typescript
// In index.ts - super naive, just passes an entity directly into the client
  await jobState.iterateEntities(
    { _type: CdnEntities.PROFILE._type },
    async (profileEntity) => {
      await client.iterateEndpoints(
        (profileEntity as unknown) as { name: string; id: string },
        ... // some other stuff here

// In client.ts - the client does the work to figure out the resourceGroupName
  public async iterateEndpoints(
    profile: { name: string; id: string },
    callback: (e: Endpoint) => void | Promise<void>,
  ): Promise<void> {
    const serviceClient = await this.getAuthenticatedServiceClient(
      CdnManagementClient,
    );
    const resourceGroup = resourceGroupName(profile.id, true)!;
    const profileName = profile.name;

    return iterateAllResources({
    ...
```

ndowmon: As for @aiwilliams' comment, the type annotations I've been adding do seem like an antipattern. We could do something like this with type annotations (assuming the `client.iterateDomainTopicSubscriptions()` interface changes to accept `{ name: string; id: string; }`, and the `resourceGroupName()` and `getEventGridDomainNameFromId()` calls are moved to inside of `client.iterateDomainTopicSubscriptions()`):

```typescript
// in converters.ts

export interface DomainTopicEntity extends Entity {
  name: string;
  id: string;
}

export function createEventGridDomainTopicEntity(
  webLinker: AzureWebLinker,
  data: DomainTopic,
): DomainTopicEntity {
    entityData: {
      source: data,
      assign: {
        name: data.name,
        type: data.type,
        ... // all of the other properties that we convert
      },
    },
  }) as DomainTopicEntity; // still type annotating but shifting left... 
}

// in index.ts
  await jobState.iterateEntities<DomainTopicEntity>( // iterateEntities already supports generics
    { _type: EventGridEntities.DOMAIN_TOPIC._type },
    async (domainTopicEntity) => { // no type annotation necessary!
      await client.iterateDomainTopicSubscriptions(
        domainTopicEntity,
        async (domainTopicSubscription) => { 
          ...
```

terraform/iam.tf
ndowmon: @KeionneDerousselle I know at one point we joked that this could be a windows issue... not sure if you have any idea whether that's credible or not. For the record, I appear to be on `terraform-provider-azurerm_v2.19.0_x5`

terraform/event-grid.tf
```diff
@@ -0,0 +1,79 @@
+variable "create_event_grid_domain" {
```
ndowmon, line 1: Nice terraform!

src/azure/utils.ts
```diff
@@ -71,3 +72,24 @@ export function resourceGroupName(
...
+ * @param id the resource id, which is expected to include the domain name
+ * @returns the resource domain name, or `undefined` when not found
+ */
+export function getEventGridDomainNameFromId(
+  id: string | undefined,
+): string | undefined {
+  let domainName: string | undefined;
+
+  if (id) {
+    const domainNameRegex = new RegExp(EVENT_GRID_DOMAIN_NAME_MATCHER, 'g');
```
ndowmon, line 87: Not a big deal - I think the `EVENT_GRID_DOMAIN_NAME_MATCHER` could probably live in this file, since it's not referenced elsewhere.

CHANGELOG.md
```diff
@@ -8,6 +8,21 @@ and this project adheres to
...
+- Added `azure_event_grid_domain|has|azure_event_grid_domain_topic`
+  relationships
+- Added `azure_event_grid_topic` entities
+- Added `azure_resource_group|has|azure_event_grid_topic` relationships
+- Added `azure_event_grid_topic_subscription` entities
+- Added `azure_event_grid_topic|has|azure_event_grid_topic_subscription`
+  relationship
+- Added `azure_event_grid_domain_topic|has|azure_event_grid_topic_subscription`
+  relationship
+
```
ndowmon, line 25: 👑

ndowmon (changes requested):
@KeionneDerousselle Awesome work on this PR! There is a _lot_ to learn in the J1 SDK and in this project in particular, you've picked up Terraform, Typescript, and the SDK impressively. Nice work.

I noted a bug in the client about `listNext` methods, which is the main request from my review. It was easy to overlook and I (or you, if you're up for it!) can work on improving the typings there so we guard from that happening in the future.

I also think it'd be great if you make calls to `resourceGroupName()` and `getEventGridDomainNameFromId()` within the `EventGridClient` methods, so that you can just pass whole entities into the iterator and keep the `fetchEventGrid<Resource>()` functions dead simple.

Overall, I'm very excited to see this PR and I think you've done a great job on your first contribution. Glad that you were able to ID & work around that confusing terraform bug and also for all your thoughts on the complex global/regional subscriptions. Let's find some time early this week to jump on a call with @aiwilliams and talk about those global/regional subscriptions & where we go from here.

src/steps/resource-manager/event-grid/client.ts
```diff
@@ -0,0 +1,169 @@
...
+      serviceClient,
+      resourceEndpoint: {
+        list: async () =>
+          serviceClient.eventSubscriptions.listByResource(
+            resourceGroupName,
+            providerNamespace,
+            type,
+            name,
+          ),
+        linkNext: serviceClient.eventSubscriptions.listByResourceNext,
```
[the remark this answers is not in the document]
ndowmon: #168

terraform/iam.tf
ndowmon: Looks like you could use `azurerm_role_definition.j1dev_subscription.role_definition_resource_id` instead of `azurerm_role_definition.j1dev_subscription.id`, they did apparently introduce a breaking change to the `.id` property: https://github.com/terraform-providers/terraform-provider-azurerm/issues/8618#issuecomment-698680205

The only problem is that `.role_definition_resource_id` is not available pre-v2.29.0, so after you make that change you'll need to lock terraform to 2.29.0 or later. https://github.com/terraform-providers/terraform-provider-azurerm/commit/0322cbc5dfb92e48d0d15aca0da844553c9443e7#diff-f1c391001620c957859ab0e8fa91084d

src/steps/resource-manager/event-grid/index.ts
KeionneDerousselle: @aiwilliams and @ndowmon, I passed in a new object to try and adhere to the interface segregation principle. I thought that sending in the entire entity was giving the client, whose responsibility is to make a call to Azure, access to way more information than it needed. In addition, the client, who doesn't even create the entity, now knows how to accept and intelligently parse an entity. It seemed like the business logic of translating and interpreting entities was done in the `index.ts` file, and the client's sole responsibility was connecting to and retrieving data from Azure.

terraform/iam.tf
KeionneDerousselle: 👍

src/steps/resource-manager/event-grid/converters.test.ts
KeionneDerousselle: 👍

src/steps/resource-manager/event-grid/client.ts
```diff
@@ -0,0 +1,169 @@
...
+      } as ListResourcesEndpoint,
+      resourceDescription: 'eventGrid.subscription',
```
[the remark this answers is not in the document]
KeionneDerousselle: Sure thing. I thought the description had to match what type being worked with (i.e. Domain Topic Subscriptions and regular Topic Subscriptions return the same type).

src/steps/resource-manager/event-grid/index.ts
```diff
@@ -0,0 +1,296 @@
...
+
+  const webLinker = createAzureWebLinker(accountEntity.defaultDomain as string);
+  const client = new EventGridClient(instance.config, logger);
+
+  await jobState.iterateEntities(
+    { _type: RESOURCE_GROUP_ENTITY._type },
+    async (resourceGroupEntity) => {
+      const { name } = resourceGroupEntity;
+      await client.iterateDomains(
+        ({ name } as unknown) as { name: string },
```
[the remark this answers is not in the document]
KeionneDerousselle: I know that the properties will line up in the end. This is the same as the conversation before. Although some of the properties on the entity line up with the dependencies the client needs to perform a call, I thought it would be good to separate those concepts. The client doesn't know about entities, or how to interpret the data on them. It just knows it needs a resource group name to perform a call. The `index.ts` does know about entities. It contains the business logic for creating them, retrieving ones it previously created, stripping, and interpreting data on them.

src/azure/utils.ts
KeionneDerousselle: 👍

src/steps/resource-manager/event-grid/converters.test.ts
```diff
@@ -0,0 +1,155 @@
...
+
+describe('createEventGridTopicSubscriptionEntity', () => {
+  test('properties transferred', () => {
+    const data: EventSubscription = {
+      id:
+        '/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/j1dev/providers/Microsoft.EventGrid/topics/j1dev-event-grid-topic/providers/Microsoft.EventGrid/eventSubscriptions/j1dev-event-grid-subscription',
+      name: 'j1dev-event-grid-subscription',
+      provisioningState: 'Succeeded',
+      type: 'Microsoft.EventGrid/eventSubscriptions',
+    };
```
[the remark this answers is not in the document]
KeionneDerousselle: I'm not sure. I will check.

src/steps/resource-manager/event-grid/index.test.ts
```diff
@@ -0,0 +1,302 @@
...
+  context.jobState.getData = jest.fn().mockResolvedValue({
+    defaultDomain: 'www.fake-domain.com',
+  });
+
+  await fetchEventGridDomainTopicSubscriptions(context);
+
+  expect(context.jobState.collectedEntities.length).toBeGreaterThan(0);
+  expect(context.jobState.collectedEntities).toMatchGraphObjectSchema({
+    _class: 'Subscription',
+    schema: {},
```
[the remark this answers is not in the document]
KeionneDerousselle: I agree. That's a good idea.

src/steps/resource-manager/event-grid/converters.test.ts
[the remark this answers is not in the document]
KeionneDerousselle: I missed some properties and will add those back.

src/steps/resource-manager/event-grid/client.ts
```diff
@@ -0,0 +1,169 @@
...
+
+    const { resourceGroupName, name } = domain;
+
+    return iterateAllResources({
+      logger: this.logger,
+      serviceClient,
+      resourceEndpoint: {
+        list: async () =>
+          serviceClient.domainTopics.listByDomain(resourceGroupName, name),
+        linkNext: serviceClient.domainTopics.listByDomainNext,
```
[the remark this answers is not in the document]
KeionneDerousselle: 👍

README.md
```diff
@@ -13,7 +13,7 @@ You may use use Node to execute directly on your machine.
 
 Prerequisites:
 
-1.  Install terraform and tfenv (see [Development](development.md))
+1.  Install terraform and tfenv (see [Development](/docs/development.md))
```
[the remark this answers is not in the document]
KeionneDerousselle: 👍

src/steps/resource-manager/event-grid/client.ts
```diff
@@ -0,0 +1,169 @@
...
+      logger: this.logger,
+      serviceClient,
+      resourceEndpoint: {
+        list: async () =>
+          serviceClient.eventSubscriptions.listByDomainTopic(
+            resourceGroupName,
+            domainName,
+            name,
+          ),
+        linkNext: serviceClient.eventSubscriptions.listByDomainTopicNext,
```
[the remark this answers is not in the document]
KeionneDerousselle: 👍

src/steps/resource-manager/event-grid/client.ts
```diff
@@ -0,0 +1,169 @@
...
+      EventGridManagementClient,
+    );
+
+    return iterateAllResources({
+      logger: this.logger,
+      serviceClient,
+      resourceEndpoint: {
+        list: async () =>
+          serviceClient.topics.listByResourceGroup(resourceGroup.name),
+        linkNext: serviceClient.topics.listByResourceGroupNext,
```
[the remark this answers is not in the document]
KeionneDerousselle: 👍

src/steps/resource-manager/event-grid/index.ts
aiwilliams: I agree with your goal to avoid passing too much data to a function! It helps communicate what the function will do when the types it receives are stated in the most narrow interface necessary/possible. However, it should be possible to pass in anything that "talks like a duck".

I also agree that the clients have a responsibility only to interact with the provider APIs, and we should strive to avoid knowing about `Entity` and other interfaces/types. This helps to make it clear that the clients are focused on fetching data from the provider, not on interpreting the world of the code that uses the client. The client then becomes unnecessarily coupled to types it doesn't actually care about.

That said, the client _is_ aware of the data structures of the provider, so it should be able to extract things from the `id` values etc.

Anyway, just talked with @ndowmon and understand this has been worked out, seeing what I was going to communicate here in https://github.com/JupiterOne/graph-azure/pull/166/commits/caf906e0393c2b08033f058e6e0c8f70156ebd0a 😄

ndowmon (approved):
Thanks for the changes! I really like the patterns you've pushed forward in this PR, including the simplified interfaces to the client's `iterateResource()` functions.

PR #170, opened 2020-10-02
Feature/azure redis cache

KeionneDerousselle:
# Acceptance Criteria
Map the entities and relationships from Azure Cache for Redis.

# What Changes?
1. Added the **Redis Cache** entity
    A Redis Cache is provided by Azure Cache for Redis. It is an in-memory data store based on Redis. https://docs.microsoft.com/en-us/azure/azure-cache-for-redis/cache-overview
https://docs.microsoft.com/en-us/azure/architecture/patterns/cache-aside

2. Added the **Redis Firewall Rule** entity
    Firewall Rules for a Redis Cache define an IP address range describing which IPs can connect to the cache. https://docs.microsoft.com/en-us/azure/azure-cache-for-redis/cache-configure#firewall

3. Added the **Redis Cache** has **Redis Firewall Rule** relationship

4. Added the **Redis Cache** connects **Redis Cache** relationship
    Two Redis Caches can be connected or linked via a Redis Linked Server. A Linked Server is a geo-replication that links two Premium tier Azure Cache for Redis instances. One cache is the primary linked cache, and the other is selected as the secondary linked cache. The secondary cache is read-only and when data is written to the primary cache, it is replicated to the secondary cache. Azure Linked Servers are designed to be a disaster-recovery solution. https://docs.microsoft.com/en-us/azure/azure-cache-for-redis/cache-how-to-geo-replication

    While a Linked Server does have some properties like an entity would, it mostly describes a relationship between two existing Redis Caches. It is for this reason that I decided to make Linked Servers a `connects` relationship between two Redis Caches. This relationship has extra properties to capture any additional data provided by the Linked Server.

# Graph
redis-graph

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ and this project adheres to
 - Added `azure_resource_group|has|azure_redis_cache` relationships
 - Added `azure_redis_firewall_rule` entities
 - Added `azure_redis_cache|has|azure_redis_firewall_rule` relationships
+- Added `azure_redis_cache|connects|azure_redis_cache` relationships
 
 ## 5.3.1 - 2020-09-21
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,8 @@ and this project adheres to
 - Added `azure_batch_account|has|azure_batch_certificate` relationships
 - Added `azure_redis_cache` entities
 - Added `azure_resource_group|has|azure_redis_cache` relationships
+- Added `azure_redis_firewall_rule` entities
+- Added `azure_redis_cache|has|azure_redis_firewall_rule` relationships
 
 ## 5.3.1 - 2020-09-21
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,8 @@ and this project adheres to
 - Added `azure_batch_account|has|azure_batch_application` relationships
 - Added `azure_batch_certificate` entities
 - Added `azure_batch_account|has|azure_batch_certificate` relationships
+- Added `azure_redis_cache` entities
+- Added `azure_resource_group|has|azure_redis_cache` relationships
 
 ## 5.3.1 - 2020-09-21
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -77,7 +77,7 @@ export const invocationConfig: IntegrationInvocationConfig<IntegrationConfig> = 
     ...batchSteps,
     ...redisCacheSteps,
 
-    // NOTE: Because any resource in Azure could be an Event Grid Topic, this step should be executed last.
+    // NOTE: Because any resource in Azure could be an Event Grid Topic, this step should be executed last. See SDK #326: https://github.com/JupiterOne/sdk/issues/326
     // This will ensure that other resources that an organization has can be tracked as 'topics' so that we can associate Event Grid Topic Subscriptions to them.
     ...eventGridSteps,
   ],
--- a/src/index.ts
+++ b/src/index.ts
@@ -21,6 +21,7 @@ import { containerRegistrySteps } from './steps/resource-manager/container-regis
 import { serviceBusSteps } from './steps/resource-manager/service-bus';
 import { cdnSteps } from './steps/resource-manager/cdn';
 import { batchSteps } from './steps/resource-manager/batch';
+import { redisCacheSteps } from './steps/resource-manager/redis-cache';
 import { eventGridSteps } from './steps/resource-manager/event-grid';
 
 export function hasSubscriptionId(config: IntegrationConfig): boolean {
@@ -74,6 +75,10 @@ export const invocationConfig: IntegrationInvocationConfig<IntegrationConfig> = 
     ...serviceBusSteps,
     ...cdnSteps,
     ...batchSteps,
+    ...redisCacheSteps,
+
+    // NOTE: Because any resource in Azure could be an Event Grid Topic, this step should be executed last.
+    // This will ensure that other resources that an organization has can be tracked as 'topics' so that we can associate Event Grid Topic Subscriptions to them.
     ...eventGridSteps,
   ],
 };
--- a/src/steps/resource-manager/redis-cache/constants.ts
+++ b/src/steps/resource-manager/redis-cache/constants.ts
@@ -6,6 +6,7 @@ import { createResourceGroupResourceRelationshipMetadata } from '../utils/create
 
 export const STEP_RM_REDIS_CACHES = 'rm-redis-caches';
 export const STEP_RM_REDIS_FIREWALL_RULES = 'rm-redis-firewall-rules';
+export const STEP_RM_REDIS_LINKED_SERVERS = 'rm-redis-linked-servers';
 
 export const RedisCacheEntities = {
   CACHE: {
@@ -35,5 +36,16 @@ export const RedisCacheRelationships = {
     sourceType: RedisCacheEntities.CACHE._type,
     _class: RelationshipClass.HAS,
     targetType: RedisCacheEntities.FIREWALL_RULE._type,
+  },
+
+  REDIS_CACHE_IS_LINKED_TO_REDIS_CACHE: {
+    _type: generateRelationshipType(
+      RelationshipClass.CONNECTS,
+      RedisCacheEntities.CACHE._type,
+      RedisCacheEntities.CACHE._type,
+    ),
+    sourceType: RedisCacheEntities.CACHE._type,
+    _class: RelationshipClass.CONNECTS,
+    targetType: RedisCacheEntities.CACHE._type,
   },
 };
--- a/src/steps/resource-manager/redis-cache/constants.ts
+++ b/src/steps/resource-manager/redis-cache/constants.ts
@@ -1,0 +1,17 @@
+import { createResourceGroupResourceRelationshipMetadata } from '../utils/createResourceGroupResourceRelationship';
+
+export const STEP_RM_REDIS_CACHES = 'rm-redis-caches';
+
+export const RedisCacheEntities = {
+  CACHE: {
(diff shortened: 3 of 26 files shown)


terraform/redis-cache.tf
```diff
@@ -30,4 +37,63 @@ resource "azurerm_redis_firewall_rule" "j1dev" {
...
+}
+
+resource "azurerm_redis_linked_server" "j1dev-redis-linked-server" {
+  count                       = local.redis_cache_count
+  target_redis_cache_name     = azurerm_redis_cache.j1dev-primary-redis-cache[count.index].name
+  resource_group_name         = azurerm_redis_cache.j1dev-primary-redis-cache[count.index].resource_group_name
+  linked_redis_cache_id       = azurerm_redis_cache.j1dev-secondary-redis-cache[count.index].id
+  linked_redis_cache_location = azurerm_redis_cache.j1dev-secondary-redis-cache[count.index].location
+  server_role                 = "Secondary"
+  depends_on                  = [azurerm_redis_cache.j1dev-primary-redis-cache, azurerm_redis_cache.j1dev-secondary-redis-cache]
```
ndowmon, line 98: What's the purpose of these `depends_on` properties? I would have assumed that terraform could infer this, considering it explicitly references these resources on lines 93-96. Were you experiencing any weird behavior that caused you to explicitly set `depends_on`?

ndowmon: Same goes for the other instances in this file.

CHANGELOG.md
```diff
@@ -34,6 +34,7 @@ and this project adheres to
 - Added `azure_resource_group|has|azure_redis_cache` relationships
 - Added `azure_redis_firewall_rule` entities
 - Added `azure_redis_cache|has|azure_redis_firewall_rule` relationships
+- Added `azure_redis_cache|connects|azure_redis_cache` relationships
```
ndowmon, line 37: Really like your choice to create a relationship here. Adam has used the term "Associative Records" to describe a relationship that's defined by an API response. I think these `linkedServer` responses are a great example of an "Associative Record".

src/steps/resource-manager/redis-cache/index.ts
```diff
@@ -63,4 +105,16 @@ export const redisCacheSteps: Step<
...
     executionHandler: fetchRedisCaches,
   },
+  {
+    id: STEP_RM_REDIS_FIREWALL_RULES,
+    name: 'Redis Firewall Rules',
+    entities: [RedisCacheEntities.FIREWALL_RULE],
+    relationships: [RedisCacheRelationships.REDIS_CACHE_HAS_FIREWALL_RULE],
+    dependsOn: [
+      STEP_AD_ACCOUNT,
+      STEP_RM_RESOURCES_RESOURCE_GROUPS,
```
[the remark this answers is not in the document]
ndowmon: Same goes for the `STEP_RM_RESOURCES_RESOURCE_GROUPS` dependency in `STEP_RM_REDIS_LINKED_SERVERS`.

terraform/redis-cache.tf
```diff
@@ -0,0 +1,99 @@
...
+    maxmemory_delta    = 2
+    maxmemory_policy   = "allkeys-lru"
+  }
+
+  depends_on = [azurerm_resource_group.j1dev]
+}
+
+resource "azurerm_redis_firewall_rule" "j1dev" {
+  count = local.redis_cache_count
+  # NOTE: You cannot choose a name with hyphens, only alphanumeric characters and underscores are allowed
```
ndowmon, line 34: So _weird_ that some resources can _only_ have hyphens, and others can _only_ have underscores... some of Azure's inconsistencies I don't think we'll ever really understand.

src/steps/resource-manager/redis-cache/index.ts
```diff
@@ -0,0 +1,186 @@
...
+    { _type: RedisCacheEntities.CACHE._type },
+    async (redisCacheEntity) => {
+      const { id, name } = redisCacheEntity;
+      const resourceGroup = resourceGroupName(id, true)!;
+
+      await client.iterateLinkedServers(
+        { resourceGroupName: resourceGroup, redisCacheName: name as string },
+        async (linkedServer) => {
+          const { linkedRedisCacheId } = linkedServer;
+          if (cacheIsLinkedToSecondaryCache(linkedServer)) {
```
ndowmon, line 123: What happens if this is false? Looks like we just skip the record, but is there any information that would still be useful to add to the graph?

ndowmon (changes requested):
Thanks for adding this service @KeionneDerousselle, this looks great! I'm glad you picked up on the "Associative Relationship" pattern and created relationships for the linked servers.

I added a few suggestions & suggestions when you've got a chance to review.

src/steps/resource-manager/redis-cache/index.ts
KeionneDerousselle: This is to reduce duplication. The client call returns the linked server info in both directions. Meaning when we ask it for linked servers for `cache1` it returns that `cache1` is linked to `cache2`. When we ask if for linked servers for `cache2`, it returns that `cache2` is linked to `cache1`. We only want to record that relationship one time.

terraform/redis-cache.tf
KeionneDerousselle: Isn't it though?! Gotta love Azure.

src/steps/resource-manager/redis-cache/index.ts
KeionneDerousselle: I'll add a comment in the code to explain this.

terraform/redis-cache.tf
KeionneDerousselle: I added these dependency declarations because, without them, terraform will attempt to remove the caches before destroying the link between them. Then it will complain that you cannot remove a Redis Cache that is linked to another cache. Both caches should be deleted only after the linked server is deleted. For some reason, terraform wasn't able to do this until I explicitly told it that the linked server was dependent on the two caches and so on.

KeionneDerousselle: I will add a comment in the code to clarify this.

terraform/redis-cache.tf
```diff
@@ -30,4 +37,63 @@ resource "azurerm_redis_firewall_rule" "j1dev" {
...
+
+  depends_on = [azurerm_resource_group.j1dev-secondary-redis-cache-resource-group]
+}
+
+resource "azurerm_redis_linked_server" "j1dev-redis-linked-server" {
+  count                       = local.redis_cache_count
+  target_redis_cache_name     = azurerm_redis_cache.j1dev-primary-redis-cache[count.index].name
+  resource_group_name         = azurerm_redis_cache.j1dev-primary-redis-cache[count.index].resource_group_name
+  linked_redis_cache_id       = azurerm_redis_cache.j1dev-secondary-redis-cache[count.index].id
+  linked_redis_cache_location = azurerm_redis_cache.j1dev-secondary-redis-cache[count.index].location
```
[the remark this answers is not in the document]
KeionneDerousselle: 👍

src/steps/resource-manager/redis-cache/constants.ts
```diff
@@ -1,17 +1,39 @@
...
 export const RedisCacheEntities = {
   CACHE: {
     _type: 'azure_redis_cache',
     _class: ['Database', 'DataStore', 'Cluster'],
     resourceName: '[RM] Redis Cache',
   },
+
+  FIREWALL_RULE: {
+    _type: 'azure_firewall_rule',
+    _class: ['Rule'],
```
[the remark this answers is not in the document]
KeionneDerousselle: 👍

src/steps/resource-manager/redis-cache/index.ts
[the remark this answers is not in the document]
KeionneDerousselle: That makes sense. 👍

src/index.ts
```diff
@@ -74,6 +75,10 @@ export const invocationConfig: IntegrationInvocationConfig<IntegrationConfig> =
     ...serviceBusSteps,
     ...cdnSteps,
     ...batchSteps,
+    ...redisCacheSteps,
+
+    // NOTE: Because any resource in Azure could be an Event Grid Topic, this step should be executed last.
+    // This will ensure that other resources that an organization has can be tracked as 'topics' so that we can associate Event Grid Topic Subscriptions to them.
```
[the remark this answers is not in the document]
KeionneDerousselle: 👍

src/steps/resource-manager/redis-cache/index.ts
```diff
@@ -120,6 +120,19 @@ export async function fetchRedisLinkedServers(
...
+           * Meaning, when we ask it for linked servers for cache1, it returns that cache1 is linked to cache2.
+           * When we ask the client for linked servers for cache2, it returns that cache2 is linked to cache1.
+           * We only want to record that relationship one time.
+           * We can use a property that the linkedServer has called serverRole to determine whether or not we will record the linked server relationship.
+           * A serverRole can only ever be 'Primary' or 'Secondary'. A cache cannot be linked to more than one cache.
+           * A cache will only ever be the 'Primary' or 'Secondary' cache in a linked server relationship.
+           * This means we only need to record the times when the linkedServer we found for the current redisCacheEntity is the 'Secondary' cache.
+           * When we do this, it means that the current redisCacheEntity in the iteration is the 'Primary' cache in the linked server relationship,
+           * and the linked server we found is the the 'Secondary' cache in the linked server relationship.
+           */
```
ndowmon, line 135: Sooooo valuable 💯

ndowmon (approved):
Great! Thanks for all your work here!

src/steps/resource-manager/redis-cache/index.ts
KeionneDerousselle: Np :)

PR #171, opened 2020-10-05
Feature/azure container instance

KeionneDerousselle:
# Acceptance Criteria
Map the entities and relationships from Azure Container Instance.

# What Changed?
1. Added the **Container Group** entity
    A container group is a collection of containers that get scheduled on the same host machine. The containers in a container group share a lifecycle, resources, local network, and storage volumes. 

    https://docs.microsoft.com/en-us/azure/container-instances/container-instances-container-groups#storage
    https://docs.microsoft.com/en-us/azure/container-instances/container-instances-overview

2. Added the **Container** entity
    The running containerized instance of the user's application.

3. Added the **Volume** entity
    By default, Azure Container Instances are stateless. You must mount a volume from an external store to persist state beyond the lifetime of the container.
    https://docs.microsoft.com/en-us/azure/container-instances/container-instances-volume-azure-files

4. Added the **Resource Group** has **Container Group** relationship

5. Added the **Container Group** has **Container** relationship

6. Added the **Container Group** has **Volume** relationship

7. Added the **Container** connects **Volume** relationship

# Graph
container-instance


src/steps/resource-manager/container-instance/contants.ts
```diff
@@ -0,0 +1,66 @@
...
+  },
+
+  CONTAINER: {
+    _type: 'azure_container',
+    _class: ['Container'],
+    resourceName: '[RM] Container',
+  },
+
+  VOLUME: {
+    _type: 'azure_volume',
```
ndowmon, line 23: I'm torn wondering whether this should be `azure_volume` or `azure_container_volume`... not sure if this volume is characteristically different than other volumes that could exist in azure. @aiwilliams any thoughts?

src/steps/resource-manager/container-instance/converters.test.ts
```diff
@@ -0,0 +1,110 @@
...
+describe('createVolumeEntity', () => {
+  test('properties transferred', () => {
+    const data: VolumeWithId = {
+      id:
+        '/subscriptions/40474ebe-55a2-4071-8fa8-b610acdd8e56/resourceGroups/j1dev/providers/Microsoft.ContainerInstance/containerGroups/j1dev-container-group/volumes/nginx-mount',
+      name: 'nginx-mount',
+      azureFile: {
+        shareName: 'j1dev',
+        readOnly: false,
+        storageAccountName: 'keionnedj1dev',
```
ndowmon, line 98: It seems like there should also be a `azure_storage_share--HAS-->azure_volume` relationship here?

ndowmon (changes requested):
@KeionneDerousselle Looks good! There's one additional relationship that I think needs to be created, I mentioned in a comment. I also posted some entity/relationship class changes.

src/steps/resource-manager/container-instance/contants.ts
KeionneDerousselle: That makes sense. I can make that change.

src/steps/resource-manager/container-instance/converters.test.ts
KeionneDerousselle: I thought that volume was using the storage share as its actual mechanism for storing. Would it make sense to say that the volume uses the storage share?

src/steps/resource-manager/container-instance/contants.ts
```diff
@@ -0,0 +1,66 @@
...
+  },
+
+  CONTAINER_CONNECTS_VOLUME: {
+    _type: generateRelationshipType(
+      RelationshipClass.CONNECTS,
+      ContainerInstanceEntities.CONTAINER,
+      ContainerInstanceEntities.VOLUME,
+    ),
+    sourceType: ContainerInstanceEntities.CONTAINER._type,
+    _class: RelationshipClass.CONNECTS,
```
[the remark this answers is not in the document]
KeionneDerousselle: 👍

src/steps/resource-manager/container-instance/converters.test.ts
ndowmon: Hm, that's a good point. Yeah, I think that the relationship you proposed makes sense.

KeionneDerousselle: That might take me a while, the volumes returned as an array on the container group don't have an id. They only come back with a share name and the storage account name. I'll have to try to derive an id to find it in the tracked entities to create a relationship between the two.

KeionneDerousselle: Also, volumes don't always use Azure File Shares. They can use Secrets, and empty directory, or a GitHub repo.

https://docs.microsoft.com/en-us/azure/container-instances/container-instances-container-groups#storage

Should I create entities and relationships for that as well?

ndowmon: You can do something like this in `index.ts`-

```typescript
            container.volumeMounts?.forEach(async (volumeMount) => {
              const volumeKey = `${id}/volumes/${volumeMount.name}`;
              // Find the previously tracked volume so that we can record a relationship between the container and the volume
              const connectedVolumeEntity = await jobState.findEntity(
                volumeKey,
              );

              if (volumeMount.azureFile?.shareName !== undefined) {
                let storageShare: Entity;
                await jobState.iterateEntities({ _type: STORAGE_FILE_SHARE_ENTITY_METADATA._type },
                  (storageFileShare) => {
                    const fileShareRegex = 
                    const match = "/subscriptions/[^/]+/resource[G|g]roups/[^/]+/providers/Microsoft.Storage/storageAccounts/([^/]+)/fileServices/default/shares/([^/]+)"; 
                    const match = storageFileShare.id.match(fleShareRegex);
                    if (match) {
                      const [ _, storageAccountName, storageShareName ] = match;
                      if (storageAccountName === volumeMount.azureFile.shareName && storageShareName === volumeMount.azureFile.shareName) {
                        storageShare = storageFileShare;
                      }
                    }
                  }
                )
              }

              if (connectedVolumeEntity) {
                  createDirectRelationship({
                    _class: RelationshipClass.USES,
                    from: containerEntity,
                    to: connectedVolumeEntity,
                    properties: volumeMount,
                  }),
                );
              }
            });
```

src/steps/resource-manager/container-instance/contants.ts
aiwilliams: The AWS integration has `aws_ebs_volume` for all block storage devices, no other `_type` value with "volume" in it. So, fwiw, seems reasonable to me to have all Azure storage volumes be a single type unless it can be argued that they really are very different things when provisioned/used by a container.

src/steps/resource-manager/container-instance/converters.test.ts
aiwilliams: If you do `container.volumeMounts?.forEach(async (volumeMount) => {`, you'd have to `await` the `forEach`, which does not return a `Promise`. I tend to just use a `for/of` loop, like `for (const volumeMount of (container.volumeMounts? || []))` when there'll need to be an `async ()` in the loop.

src/steps/resource-manager/container-instance/types.ts
```diff
@@ -0,0 +1,4 @@
+import { Container, Volume } from '@azure/arm-containerinstance/esm/models';
+
+export type ContainerWithId = Container & { id: string };
+export type VolumeWithId = Volume & { id: string };
```
aiwilliams, line 4: I think it would be nice to have comments that are in the code/functions to move to these types. This will allow folks to see the purpose of these types here, and well as when hovering over the types wherever they are referenced in the functions that receive them.

src/steps/resource-manager/container-instance/index.ts
```diff
@@ -0,0 +1,158 @@
...
+                to: containerEntity,
+              }),
+            );
+
+            /**
+             * NOTE: If the Container is connected to a Volume via a volumeMount,
+             * we want to record that relationship.
+             */
+            container.volumeMounts?.forEach(async (volumeMount) => {
+              const volumeKey = `${id}/volumes/${volumeMount.name}`;
```
aiwilliams, line 114: Hmm, these really are looking like volumes that exist solely for the container (`azure_container_volume`). Is there a way to view these volumes apart from the container that defines them?

src/steps/resource-manager/container-instance/index.ts
```diff
@@ -0,0 +1,158 @@
...
+           * NOTE: There is not call/function in the Container Instance Management Client to get containers by container group.
+           * Instead, containers are returned in an array on the container group.
+           * Iterating over them in memory shouldn't cause an issue, because there is a limit set at 60 containers per container group.
+           * https://docs.microsoft.com/en-us/azure/container-instances/container-instances-quotas
+           */
+          containerGroup.containers.forEach(async (container) => {
+            /**
+             * NOTE: Containers in Azure Container Groups do not have an id. We are creating one by adding /containers/${containerName} at the end of the Azure Container Group Id
+             * The result should look like /subscriptions/${subscriptionId}/resourceGroups/${resourceGroup}/providers/Microsoft.ContainerInstance/containerGroups/${containerGroup}/containers/${containerName}
+             */
```
aiwilliams, line 94: This is the commentary that I think could be moved to the `ContainerWithId` type, which the `createContainerEntity` function receives. Perhaps there may also need to be docs on the `createContainerEntity` function. In any case, the point is, the commentary goes from being in this one place where the type and function are used to being on the type/function no matter where it ends up being used, and the editors are great with onHover doc visualization!

src/steps/resource-manager/container-instance/index.ts
```diff
@@ -0,0 +1,158 @@
...
+            );
+          });
+
+          /**
+           * NOTE: There is not call/function in the Container Instance Management Client to get containers by container group.
+           * Instead, containers are returned in an array on the container group.
+           * Iterating over them in memory shouldn't cause an issue, because there is a limit set at 60 containers per container group.
+           * https://docs.microsoft.com/en-us/azure/container-instances/container-instances-quotas
+           */
+          containerGroup.containers.forEach(async (container) => {
```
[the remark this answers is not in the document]
KeionneDerousselle: I just didn't know this. You aren't misunderstanding anything lol, I was. I will fix this.

src/steps/resource-manager/container-instance/client.test.ts
```diff
@@ -0,0 +1,106 @@
...
+                mountPath: '/etc/test_mount',
+                name: 'nginx-mount',
+                readOnly: false,
+              },
+            ],
+          },
+          {
+            command: [],
+            environmentVariables: [],
+            image: 'microsoft/aci-helloworld:latest',
```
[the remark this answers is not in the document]
KeionneDerousselle: How can we tell, without an Id, whether the image is internal to the company (already ingested and tracked) or external (like the one in this example/line 68)?

src/steps/resource-manager/container-instance/__snapshots__/converters.test.ts.snap
```diff
@@ -0,0 +1,160 @@
...
+        "azureFile": Object {
+          "readOnly": false,
+          "shareName": "j1dev",
+          "storageAccountName": "keionnedj1dev",
+        },
+        "name": "nginx-mount",
+      },
+    },
+  ],
+  "_type": "azure_container_volume",
```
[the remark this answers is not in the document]
KeionneDerousselle: I'm not sure. I don't think the volume we are representing is available outside of the container. It's just a way to map a drive to some external existing data source, I think. I might be misunderstanding something. Let's get on a call to discuss.

src/steps/resource-manager/container-instance/contants.ts
KeionneDerousselle: We discussed more about how the structure of data returns for Azure and decided that we would continue with the relationships we have. We discussed potentially removing the Container Volume entity in the future and linking the Container Group directly to the file share. We discussed moving that Volume data as properties on the Container Group uses File Share relationship. We would add the Container uses File Share relationship (essentially removing the Container Volume entity all together) and move the `volumeMount` data to the Container users File Share relationship.

src/steps/resource-manager/container-instance/__snapshots__/converters.test.ts.snap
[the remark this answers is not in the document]
KeionneDerousselle: We discussed more about how the structure of data returns for Azure and decided that we would continue with the relationships we have. We discussed potentially removing the Container Volume entity in the future and linking the Container Group directly to the file share. We discussed moving that Volume data as properties on the Container Group uses File Share relationship. We would add the Container uses File Share relationship (essentially removing the Container Volume entity altogether) and move the volumeMount data to the Container users File Share relationship.

src/steps/resource-manager/container-instance/client.test.ts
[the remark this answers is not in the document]
KeionneDerousselle: We discussed adding this later because we are not ingesting images at the moment. We also don't know how Azure treats images. (Do image names refer to an image registered with Azure or in Docker?)

src/steps/resource-manager/container-instance/index.ts
KeionneDerousselle: We discussed more about how the structure of data returns for Azure and decided that we would continue with the relationships we have. We discussed potentially removing the Container Volume entity in the future and linking the Container Group directly to the file share. We discussed moving that Volume data as properties on the Container Group uses File Share relationship. We would add the Container uses File Share relationship (essentially removing the Container Volume entity all together) and move the volumeMount data to the Container users File Share relationship.

ndowmon (approved):

aiwilliams (approved):
Thanks @KeionneDerousselle & @ndowmon !

PR #176, opened 2020-10-13
Do not throw if createResourceGroupResourceRelationship does not find…

### LINKED ISSUE #175: Steps should not throw if they cannot match a resource group ID
The azure integration throws on many steps with 

```
Could not find the resource group "/subscriptions/<subscription-id>/resourceGroups/<resourceGroup-id>" in this subscription.
```

This causes steps to not complete and resources to not be ingested.

ndowmon:
… the target resource group

Fixes #175

### DIFF
--- a/src/steps/resource-manager/compute/index.ts
+++ b/src/steps/resource-manager/compute/index.ts
@@ -47,11 +47,9 @@ export async function fetchVirtualMachines(
     const virtualMachineEntity = createVirtualMachineEntity(webLinker, vm);
     await jobState.addEntity(virtualMachineEntity);
 
-    await jobState.addRelationship(
-      await createResourceGroupResourceRelationship(
-        executionContext,
-        virtualMachineEntity,
-      ),
+    await createResourceGroupResourceRelationship(
+      executionContext,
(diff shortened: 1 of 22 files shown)


src/steps/resource-manager/compute/index.ts
```diff
@@ -97,12 +93,7 @@ export async function fetchVirtualMachineDisks(
     const diskEntity = createDiskEntity(webLinker, data);
     await jobState.addEntity(diskEntity);
 
-    await jobState.addRelationship(
-      await createResourceGroupResourceRelationship(
-        executionContext,
-        diskEntity,
-      ),
-    );
+    await createResourceGroupResourceRelationship(executionContext, diskEntity);
```
a-u-h-g, line 96: Any reason for the one line syntax choice compared to the other changes?

a-u-h-g (approved):

src/steps/resource-manager/compute/index.ts
ndowmon: Good catch - The linter probably did that. These repositories have a mind of their own 👾

PR #191, opened 2020-10-29
Feature/azure policy

KeionneDerousselle:
# Acceptance Criteria
Ingest Azure Policy to ensure we have enough data to answer the following CIS Benchmark questions:

- 2.3 Ensure ASC Default policy setting "Monitor System Updates" is not "Disabled" (Scored)
    - Ensure the ‘properties.parameters.systemUpdatesMonitoringEffect.value’ on ‘policyAssignments/SecurityCenterBuiltIn’ is not set to Disabled or empty

- 2.4 Ensure ASC Default policy setting "Monitor OS Vulnerabilities" is not "Disabled" (Scored)
    - Ensure the ‘properties.parameters.systemConfigurationsMonitoringEffect.value’ on ‘policyAssignments/SecurityCenterBuiltIn’ is not set to Disabled or empty

- 2.5 Ensure ASC Default policy setting "Monitor Endpoint Protection" is not "Disabled" (Scored)
    - Ensure the ‘properties.parameters.endpointProtectionMonitoringEffect.value’ on ‘policyAssignments/SecurityCenterBuiltIn’ is not set to Disabled or empty

- 2.6 Ensure ASC Default policy setting "Monitor Disk Encryption" is not "Disabled" (Scored)
    - Ensure the ‘properties.parameters.diskEncryptionMonitoringEffect.value’ on ‘policyAssignments/SecurityCenterBuiltIn’ is not set to Disabled or empty

- 2.7 Ensure ASC Default policy setting "Monitor Network Security Groups" is not "Disabled" (Scored)
    - Ensure the ‘properties.parameters.networkSecurityGroupsMonitoringEffect.value’ on ‘policyAssignments/SecurityCenterBuiltIn’ is not set to Disabled or empty

- 2.8 Ensure ASC Default policy setting "Monitor Web Application Firewall" is not "Disabled" (Scored)
    - Ensure the ‘properties.parameters.webApplicationFirewallMonitoringEffect.value’ on ‘policyAssignments/SecurityCenterBuiltIn’ is not set to Disabled or empty

- 2.9 Ensure ASC Default policy setting "Enable Next Generation Firewall(NGFW) Monitoring" is not "Disabled" (Scored)
    - Ensure the ‘properties.parameters.nextGenerationFirewallMonitoringEffect.value’ on ‘policyAssignments/SecurityCenterBuiltIn’ is not set to Disabled or empty

- 2.10 Ensure ASC Default policy setting "Monitor Vulnerability Assessment" is not "Disabled" (Scored)
    - Ensure the ‘properties.parameters.vulnerabilityAssesmentMonitoringEffect.value’ on ‘policyAssignments/SecurityCenterBuiltIn’ is not set to Disabled or empty

- 2.11 Ensure ASC Default policy setting "Monitor Storage Blob Encryption" is not "Disabled" (Scored)
    - Ensure the ‘properties.parameters.storageEncryptionMonitoringEffect.value’ on ‘policyAssignments/SecurityCenterBuiltIn’ is not set to Disabled or empty

- 2.12 Ensure ASC Default policy setting "Monitor JIT Network Access" is not "Disabled" (Scored)
    - Ensure the ‘properties.parameters.jitNetworkAccessMonitoringEffect.value’ on ‘policyAssignments/SecurityCenterBuiltIn’ is not set to Disabled or empty

- 2.13 Ensure ASC Default policy setting "Monitor Adaptive Application Whitelisting" is not "Disabled" (Scored)
    - Ensure the ‘properties.parameters.adaptiveApplicationControlsMonitoringEffect.value’ on ‘policyAssignments/SecurityCenterBuiltIn’ is not set to Disabled or empty

- 2.14 Ensure ASC Default policy setting "Monitor SQL Auditing" is not "Disabled" (Scored)
    - Ensure the ‘properties.parameters.sqlAuditingMonitoringEffect.value’ on ‘policyAssignments/SecurityCenterBuiltIn’ is not set to Disabled or empty

- 2.15 Ensure ASC Default policy setting "Monitor SQL Encryption" is not "Disabled" (Scored)
    - Ensure the ‘properties.parameters.sqlEncryptionMonitoringEffect.value’ on ‘policyAssignments/SecurityCenterBuiltIn’ is not set to Disabled or empty

# What Changed?
1. Added the **Policy Assignment** entity, of type **azure_policy_assignment**, and schema **ControlPolicy**.
    
    An Azure Policy Assignment contains one or more Azure Policy Definitions that are assigned to a scope (Azure Management Group, Azure Subscription, Azure Resource Group, or Azure Resource).  Assignments are, by default applied to all child resources in a specific scope. Assignments define a set of parameters to be passed into Policy Definitions.

2. Added the **ANY_SCOPE** has **Policy Assignment** relationship
     
    Management Groups, Subscriptions, Resource Groups, and Resources can all have Policy Assignments. We currently don't ingest Management Groups, but we do ingest Subscriptions and Resource Groups. We've chosen, instead, to represent the relationship as ANY_SCOPE has a Policy Assignment. This is because we want the relationship metadata to line up with the actual ingested relationships. If we say that we expect ANY_SCOPE_had_policy_assignment and instead generate that an azure_storage_account_has_policy_assignment, it might cause issues down the road.

# Decisions Made
- If we ingest a **Policy Assignment** that has a known scope, we will establish a ‘has’ relationship between that scope and the **Policy Assignment**. 

    For example, if the Policy Assignment belongs to a known Subscription, Resource Group, or Resource (such as a database, VM, cache) that has already been ingested, we will add the X has Policy Assignment relationship. 

- If the scope is unknown, not ingested, the **Policy Assignment** entity will be created, but not related to any other entities.

- We will flatten any properties on the **Policy Assignment** entity using the dot notation for the key name. If a key is ‘value’, we will omit it. 

- Because a **Policy Assignment** can be associated with any Azure Resource, the **Policy Assignment** step will be listed amongst the last steps to execute.

# Out of Scope
- Defining the **Azure Policy Definition** entity

- Defining the **Azure Policy Assignment** has **Azure Policy Definition** relationship

- Establishing a relationship between **Azure Policy Assignments** and the **Azure 
Management Group** they belong to. We currently are not ingesting **Azure Management Groups**. 

# Graph
Screen Shot 2020-10-29 at 4 32 43 PM


src/steps/resource-manager/policy/__snapshots__/converters.test.ts.snap
```diff
@@ -0,0 +1,62 @@
...
+  "createdOn": undefined,
+  "description": "This is the default set of policies monitored by Azure Security Center. It was automatically assigned as part of onboarding to Security Center. The default assignment contains only audit policies. For more information please visit https://aka.ms/ascpolicies",
+  "displayName": "ASC Default (subscription: 40474ebe-55a2-4071-8fa8-b610acdd8e56)",
+  "enforcementMode": "Default",
+  "id": "/subscriptions/40474ebe-55a2-4071-8fa8-b610acdd8e56/providers/Microsoft.Authorization/policyAssignments/SecurityCenterBuiltIn",
+  "metadata.assignedBy": "Security Center",
+  "metadata.createdBy": "488baaae-0786-48e5-aa3d-6c7c773d9557",
+  "metadata.createdOn": "2020-09-24T16:32:06.2258988Z",
+  "metadata.updatedBy": "3b3dbbda-9a3c-4b0b-b2f4-60cfeaa74233",
+  "metadata.updatedOn": "2020-10-29T17:42:36.1625526Z",
```
aiwilliams, line 52: I think we should get these moved directly to the entity: `"_class": "ControlPolicy", "createdOn": number, "createdBy": string, ... etc`. We should populate created/update timestamps from the source anytime that data is available.

Can the `createdBy` etc. values be used to build a relationship to a principal? I think we should create an Issue for those relationships to do it later and hopefully driven by other CIS Benchmark questions. 😄

src/steps/resource-manager/policy/client.test.ts
```diff
@@ -0,0 +1,115 @@
...
+
+    /**
+     * NOTE: The parameters object may return empty, instead of having the 'systemUpdatesMonitoringEffect' key in it.
+     * This will depend on whether or not you have manually changed the value in your Azure Subscription.
+     * It looks like the built in Security Center Policy Assignment parameters are ONLY returned if they have been updated.
+     * If none of the default SecurityCenterBuiltIn parameters are altered, the parameters object will come back as empty.
+     * THIS TEST WILL ONLY PASS IF THE ORIGINAL RECORDING IS NOT DELETED OR THE DEV RUNNING THE TEST HAS ALTERED THE SAME PARAMETERS
+     * TODO : Although this scenario isn't likely, we can prevent this issue in the future by using terraform to edit all the default SecurityCenterBuiltIn properties we care about
+     * If you need to edit the default SecurityCenterBuiltIn parameters:
+     * Security Center -> Security Policy -> <Your Subscription> -> Security center default policy -> View effective policy -> ASC Default (subscription: <Your Subscription Id>)
```
aiwilliams, line 51: ❤️  the details!

src/steps/resource-manager/policy/index.ts
```diff
@@ -0,0 +1,75 @@
...
+            /**
+             * NOTE: Azure Management Groups, Azure Subscriptions, Azure Resource Groups, and Azure Resources can all have Azure Policy Assignments.
+             * We currently don't ingest Management Groups, but we do ingest Subscriptions, Resource Groups, and some Resources
+             * We've chosen to represent the relationship as ANY_SCOPE has a Policy Assignment.
+             * This is because we want the relationship metadata to line up with the actual ingested relationships.
+             * If we say that we expect ANY_SCOPE_had_policy_assignment and instead generate that a azure_storage_account_has_policy_assignment,
+             * it might cause issues down the road.
+             */
+            properties: {
+              _type: 'ANY_SCOPE_has_policy_assignment',
```
aiwilliams, line 52: This could be `PolicyRelationships.ANY_RESOURCE_POLICY_ASSIGNMENT._type` to allow for a single point of modification.

src/steps/resource-manager/policy/constants.ts
```diff
@@ -0,0 +1,34 @@
...
+};
+
+export const PolicyRelationships = {
+  /**
+   * NOTE: Management Groups, Subscriptions, Resource Groups, and Resources can all have Policy Assignments
+   * We currently don't ingest Management Groups, but we do ingest Subscriptions and Resource Groups
+   * We've chosen, instead, to represent the relationship as ANY_SCOPE has a Policy Assignment.
+   * This is because we want the relationship metadata to line up with the actual ingested relationships.
+   * If we say that we expect ANY_SCOPE_had_policy_assignment and instead generate that a azure_storage_account_has_policy_assignment,
+   * it might cause issues down the road.
```
aiwilliams, line 26: How about `azure_resource_has_policy_assignment`? I don't think I've seen any uppercase chars in `_type` values, and I think there are various parts fo the system that count on snake case for `_type` values. Also, since these are vendor-specific relationships. they should have the vendor id as a prefix.

src/steps/resource-manager/policy/helpers.test.ts
```diff
@@ -0,0 +1,169 @@
+import { flatten } from './helpers';
+
+describe('flatten', () => {
```
aiwilliams, line 3: Would love to see a PR for the SDK to have this alongside `convertProperties`, perhaps as `flattenObjectPropertyValue` (for symmetry with `convertTimePropertyValue`), so that folks don't attempt to flatten everything, but use it as:

```
  displayName: data.name,
  ...flattenObjectPropertyValue(data.parameters)
```

src/steps/resource-manager/policy/helpers.ts
```diff
@@ -0,0 +1,143 @@
...
+  {
+    canConvert: (key, value, _) =>
+      typeof value === 'boolean' || typeof value === 'number',
+    convert: (key, value, _) => value,
+  },
+  {
+    canConvert: (key, value, options) =>
+      !!(typeof value === 'string' && options.parseTime && isTimeProperty(key)),
+    convert: (key, value, _) => {
+      const time = parseTimePropertyValue(value);
```
aiwilliams, line 60: 💯

aiwilliams (changes requested):

src/steps/resource-manager/policy/index.ts
```diff
@@ -0,0 +1,75 @@
...
+  const accountEntity = await jobState.getData<Entity>(ACCOUNT_ENTITY_TYPE);
+  const webLinker = createAzureWebLinker(accountEntity.defaultDomain as string);
+  const client = new AzurePolicyClient(instance.config, logger);
+
+  await client.iteratePolicyAssignments(async (policyAssignment) => {
+    const policyAssignmentEntity = createPolicyAssignmentEntity(
+      webLinker,
+      policyAssignment,
+    );
+    await jobState.addEntity(policyAssignmentEntity);
```
austinkelleher, line 31: This is pretty minor, but it's worth noting that `addEntity` now returns the entity, so you can change this to:

```ts
const policyAssignmentEntity = await jobState.addEntity(
  createPolicyAssignmentEntity(
     webLinker,
     policyAssignment,
  )
);
```

Also, I generally prefer early returns over nested conditions. I think it also helps simplify code and prevents a bunch of indentation. What do you think of changing this block of code to the following:

```ts
export async function fetchPolicyAssignments(
  executionContext: IntegrationStepContext,
): Promise<void> {
  const { instance, logger, jobState } = executionContext;
  const accountEntity = await jobState.getData<Entity>(ACCOUNT_ENTITY_TYPE);
  const webLinker = createAzureWebLinker(accountEntity.defaultDomain as string);
  const client = new AzurePolicyClient(instance.config, logger);

  await client.iteratePolicyAssignments(async (policyAssignment) => {
    const policyAssignmentEntity = await jobState.addEntity(
      createPolicyAssignmentEntity(
         webLinker,
         policyAssignment,
      )
    );

    const { scope } = policyAssignment;
    if (!scope) return;

    const scopeEntity = await jobState.findEntity(scope);
    if (!scopeEntity) return;

      createDirectRelationship({
        _class: RelationshipClass.HAS,
        from: scopeEntity,
        to: policyAssignmentEntity,
        /**
         * NOTE: Azure Management Groups, Azure Subscriptions, Azure Resource Groups, and Azure Resources can all have Azure Policy Assignments.
         * We currently don't ingest Management Groups, but we do ingest Subscriptions, Resource Groups, and some Resources
         * We've chosen to represent the relationship as ANY_SCOPE has a Policy Assignment.
         * This is because we want the relationship metadata to line up with the actual ingested relationships.
         * If we say that we expect ANY_SCOPE_had_policy_assignment and instead generate that a azure_storage_account_has_policy_assignment,
         * it might cause issues down the road.
         */
        properties: {
          _type: 'ANY_SCOPE_has_policy_assignment',
        },
      }),
    );
  });
}
```

src/steps/resource-manager/policy/helpers.ts
```diff
@@ -0,0 +1,143 @@
+import {
+  parseStringPropertyValue,
+  parseTimePropertyValue,
+} from '@jupiterone/integration-sdk-core/dist/src/data/converters';
+
+export type FlattenObjectOptions = {
```
austinkelleher, line 6: It doesn't seem like we are using any of these options right now. Is that true?

src/steps/resource-manager/policy/helpers.ts
```diff
@@ -0,0 +1,143 @@
...
+
+function getConversionStrategy(
+  key: string,
+  value: any,
+  options: FlattenObjectOptions,
+): any {
+  return conversionStrategies.find((cs) => cs.canConvert(key, value, options));
+}
+
+export function flatten(
```
austinkelleher, line 110: Thanks for thinking about this. We should consider putting this into the SDK once we've proven it out!

austinkelleher (dismissed):
Awesome! Thanks for your work on this. Looks good to me. Only left a few minor comments/questions.

src/steps/resource-manager/policy/constants.ts
ndowmon: @aiwilliams this is a convention that I used recently in the Azure `advisor` service. I chose uppercase to make these one-to-many type relationships easily differentiated in `jupiterone.md` files. Perhaps the `sourceType` property can still be uppercase for the sake of the docs, but we can do something different with the actual type (such as `azure_resource_has_policy_assignment`, as you noted).

src/steps/resource-manager/policy/client.ts
```diff
@@ -0,0 +1,34 @@
...
+      PolicyClient,
+    );
+    return iterateAllResources({
+      logger: this.logger,
+      serviceClient,
+      resourceEndpoint: {
+        list: async () => serviceClient.policyAssignments.list(),
+        listNext: async (nextPageLink) =>
+          serviceClient.policyAssignments.listNext(nextPageLink),
+      },
```
ndowmon, line 29: If there are no arguments to the `serviceClient.policyAssignments.list()` function, you can simply pass as below:

```suggestion
      resourceEndpoint: serviceClient.policyAssignment,
```

src/steps/resource-manager/policy/helpers.test.ts
ndowmon: I wonder if some of these can be added to the `ConvertPropertiesOptions`: https://github.com/JupiterOne/sdk/blob/master/packages/integration-sdk-core/src/data/converters.ts#L35

src/steps/resource-manager/policy/helpers.ts
ndowmon: It looks like these properties (or at least some of them) come from the `ConvertPropertiesOptions` interface... https://github.com/JupiterOne/sdk/blob/master/packages/integration-sdk-core/src/data/converters.ts#L35

ndowmon (commented):
This looks good! Thanks!

src/steps/resource-manager/policy/helpers.test.ts
KeionneDerousselle: https://github.com/JupiterOne/sdk/issues/369

src/steps/resource-manager/policy/helpers.ts
KeionneDerousselle: https://github.com/JupiterOne/sdk/issues/369

src/steps/resource-manager/policy/helpers.ts
KeionneDerousselle: https://github.com/JupiterOne/sdk/issues/369

src/steps/resource-manager/policy/converters.ts
```diff
@@ -12,25 +12,26 @@ export function createPolicyAssignmentEntity(
...
+  const assignedBy = metadata?.assignedBy;
+  const createdOn = parseTimePropertyValue(metadata?.createdOn);
+  const updatedOn = parseTimePropertyValue(metadata?.updatedOn);
+  const createdBy = metadata?.createdBy;
+  const updatedBy = metadata?.updatedBy;
 
   return createIntegrationEntity({
     entityData: {
-      source: data,
+      source: pa,
```
ndowmon, line 24: I think source should still be `data`, including the metadata. The metadata property ought to be available in raw data.

ndowmon (changes requested):

src/steps/resource-manager/policy/converters.ts
KeionneDerousselle: When I had that, I got "duplicate" keys. 
```
{
  'metadata.assignnedBy': '',
  'metadata.createdOn': '',
  createdOn: '',

}
```

KeionneDerousselle: Is that what we want? I thought @aiwilliams wanted us to pull those properties to the top level of the entity.

aiwilliams: I don't think we want that to happen. Is the `source.metadata` being transferred "automatically" by something in `createIntegrationEntity()`?

KeionneDerousselle: I'm not sure if assigning `source` to `data` or the `assign: { ...flatten(data), ...)` was the culprit. I thought we just wanted to specifically peel those props off and move them to the very top so I spread them on at the root level at the end of the object.

KeionneDerousselle: I originally had:
```typescript
export function createPolicyAssignmentEntity(
  webLinker: AzureWebLinker,
  data: PolicyAssignment,
): Entity {
    entityData: {
      source: data,
      assign: {
        ...flatten(data),
        _key: pa.id,
        _type: PolicyEntities.POLICY_ASSIGNMENT._type,
        _class: PolicyEntities.POLICY_ASSIGNMENT._class,
        id: pa.id,
        name: pa.displayName,
        webLink: webLinker.portalResourceUrl(pa.id),
        scope: pa.scope,
        policyDefinitionId: pa.policyDefinitionId,
      },
    },
  });
}
```
I then changed it to:
```typescript
export function createPolicyAssignmentEntity(
  webLinker: AzureWebLinker,
  data: PolicyAssignment,
): Entity {
  const { metadata, ...pa } = data;
  const assignedBy = metadata?.assignedBy;
  const createdOn = parseTimePropertyValue(metadata?.createdOn);
  const updatedOn = parseTimePropertyValue(metadata?.updatedOn);
  const createdBy = metadata?.createdBy;
  const updatedBy = metadata?.updatedBy;

    entityData: {
      source: pa, // to remove metadata
      assign: {
        ...flatten(pa), // to remove metadata
        _key: pa.id,
        _type: PolicyEntities.POLICY_ASSIGNMENT._type,
        _class: PolicyEntities.POLICY_ASSIGNMENT._class,
        id: pa.id,
        name: pa.displayName,
        webLink: webLinker.portalResourceUrl(pa.id),
        scope: pa.scope,
        policyDefinitionId: pa.policyDefinitionId,
        ...(assignedBy && { assignedBy }),
        ...(createdBy && { createdBy }),
        ...(createdOn && { createdOn }),
        ...(updatedBy && { updatedBy }),
        ...(updatedOn && { updatedOn }),
      },
    },
  });
}
```

KeionneDerousselle: Are we saying we just want to keep `source: data` to have the original data as the source and leave `assign: { ...flatten(pa), ... }`?

ndowmon: @KeionneDerousselle yes, you should just change `source: pa` back to `source: data`. The change you made for `...flatten(pa)` was the only thing that needed to change. This way, the entity properties include `createdOn`, etc instead of `metadata.createdOn`, but the `_rawData` of the entity will include `metadata: { createdOn: ... }`

ndowmon (approved):

aiwilliams (approved):

austinkelleher (approved):

PR #193, opened 2020-11-03
Feature/azure security center

KeionneDerousselle:
# Acceptance Criteria
- Ingest security contacts using “securityCenterServiceClient.securityContacts.list”
- Requirements
    - 2.16 Ensure that 'Security contact emails' is set (Scored)
    - 2.17 Ensure that security contact 'Phone number' is set (Scored)
    - 2.18 Ensure that 'Send email notification for high severity alerts' is set to 'On' (Scored)
    - 2.19 Ensure that 'Send email also to subscription owners' is set to 'On' (Scored)

# What Changed?
1. Added the **Security Center Contact** entity, of type **azure_security_center_contact**, and schema **Resource**

2. Added the **Subscription** has **Security Center Contact** relationship

# Decisions Made
- Because a **Security Center Contact** email could be the email of a non-user, comma-separated list, or distribution list, we did not attempt to relate **Security Center Contacts** to users in the Azure Subscription.

- Because the **User** data model/schema was more related to a login/authentication account, we did not use it.

- Because the **Person** data model/schema was more related to a person with a first and last name, we did not use it.

- For now, we are using the **Resource** type to describe the **Security Center Contact** entity, until a new model type is made.  **Security Center Contact** are more related to contact details (one email, list of emails, distribution list, phone number), than any one specific user or person.

# Graph
Screen Shot 2020-11-03 at 4 01 24 PM


austinkelleher (approved):
We are going to stick with the `Resource` class for now and add a `Contact` class later if we see more examples where it could be used. Looks good to me!

ndowmon (approved):
Looks good! Thanks for being patient while we got an answer to the `Resource`/`Contact` `_class` debate.

PR #199, opened 2020-12-07
Feature/diagnostic settings

KeionneDerousselle:
# Acceptance Criteria

- 5.1.5 Ensure that logging for Azure KeyVault is 'Enabled' (Automated)
    - This benchmark aims to ensure that AuditEvent logging is enabled for Key Vault instances. This ensures that interactions with Key Vaults are logged and available. This information is available in the array of LogSettings on the returned Diagnostic Setting object(s) for a Key Vault. The CIS Benchmark documentation says to make sure that ‘category’, ‘days’, and ‘storageAccountId’ are all set.

# What Changed?
- Added the **Azure Diagnostic Log Setting** entity
- Added the **Azure Diagnostic Metric Setting** entity
- Added the **Azure Resource** has **Azure Diagnostic Log Setting** relationship
- Added the **Azure Resource** had **Azure Diagnostic Metric Setting** relationship
- Added the **Azure Diagnostic Log Setting** uses **Azure Storage Account** relationship
- Added the **Azure Diagnostic Metric Setting** uses **Azure Storage Account** relationship

# Decisions Made
- Currently, Log and Metric Settings appear as arrays of objects on the **Azure Diagnostic Settings** returned for a given resource. Because querying these arrays of objects as JSON to answer the CIS benchmarks would be difficult, we decided to create each of these objects as their own entity.

- Log Settings and Metric Settings did not have ids, so we created our own to use as an entity key. There were no properties on these objects that uniquely identified them from other Log or Metrics Settings in the array. Therefore, we created a composite identifier. This identifier is made up of the id of the setting's parent object, the **Azure Diagnostic Settings**, and the properties on the Log or Metric Setting itself. 
    - The id of an Azure Diagnostic Log Setting entity should look like: `${diagnosticSettings.id}/logs/${category}/${enabled}/${retentionPolicy?.days}/${retentionPolicy?.enabled}`
    - The id of an Azure Diagnostic Metric Setting entity should look like : `${diagnosticSettings.id}/metrics/${category}/${enabled}/${timeGrain}/${retentionPolicy?.days}/${retentionPolicy?.enabled}`

- The list() function on the Diagnostic Settings endpoint does not return all Diagnostic Settings for an Azure Subscription. All list() functions require that you pass them an Id of the resource that you would like to get a list of Diagnostic Settings for. This means that you have to have the Id of the resource that you want Diagnostic Settings for before making this call. Knowing this, we decided to first get the **Azure Diagnostic Settings** for an **Azure Key Vault**. We will iterate on this later to make sure we are getting the **Azure Diagnostic Settings** for [all Azure resources that support **Diagnostic Settings**](https://docs.microsoft.com/en-us/azure/azure-monitor/platform/resource-logs-categories#microsoftsqlserversdatabases).
    - We discussed potentially making a module, change to the Client class, or baking some automated process into the architecture to collect **Diagnostic Settings** for a resource. This would be preferable to manually retrieving **Diagnostic Settings** for each resource that supports it, especially because the number of supported resources change.

# Out of Scope
- Although **Diagnostic Log Settings** can configure a **Log Analytic** workspace that logs should be stored in, we did not create a relationship between **Diagnostic Log Settings** and **Log Analytic** workspaces. This is because we are not currently ingesting **Log Analytics**.

- Although **Diagnostic Metric Settings** can configure a **Log Analytic** workspace that logs should be stored in, we did not create a relationship between **Diagnostic Metric Settings** and **Log Analytic** workspaces. This is because we are not currently ingesting **Log Analytics**.

- Although **Diagnostic Log Settings** can configure an **Event Hub** that logs should be stored in, we did not create a relationship between **Diagnostic Log Settings** and **Event Hubs**. This is because we are not currently ingesting **Event Hubs**.

- Although **Diagnostic Metric Settings** can configure an **Event Hub** that logs should be stored in, we did not create a relationship between **Diagnostic Metric Settings** and **Event Hubs**. This is because we are not currently ingesting **Event Hubs**.

- We decided to first get the **Azure Diagnostic Settings** for an **Azure Key Vault**. We will iterate on this later to make sure we are getting the **Azure Diagnostic Settings** for [all Azure resources that support **Diagnostic Settings**](https://docs.microsoft.com/en-us/azure/azure-monitor/platform/resource-logs-categories#microsoftsqlserversdatabases).

# Graph
Screen Shot 2020-12-10 at 2 53 44 PM


terraform/main.tf
```diff
@@ -1,6 +1,11 @@
 provider "azurerm" {
   version = "~>2.29.0"
-  features {}
+  features {
+    key_vault {
```
KeionneDerousselle, line 4: There was a change released to Key Vaults in Dec 2020. This change automatically enabled soft deletes of key vaults, because of this, we needed to tell terraform to restore soft-deleted key vaults that were torn down. You can currently create Key Vaults that are not setup to soft delete when destroyed, but this feature will be gone soon.

terraform/keyvault.tf
```diff
@@ -1,12 +1,12 @@
...
-
 resource "azurerm_key_vault" "j1dev" {
-  name                        = "${var.developer_id}-j1dev"
+  name                        = "${var.developer_id}1-j1dev"
   location                    = azurerm_resource_group.j1dev.location
   resource_group_name         = azurerm_resource_group.j1dev.name
   enabled_for_disk_encryption = true
   tenant_id                   = data.azurerm_client_config.current.tenant_id
-  soft_delete_enabled         = false
+  soft_delete_enabled         = true
```
KeionneDerousselle, line 8: I added this because the ability to disable soft delete will be gone soon. I did set the number of days to retain a soft deleted Key Vault to the lowest possible (7).

terraform/main.tf
```diff
@@ -1,6 +1,11 @@
...
+      recover_soft_deleted_key_vaults = true
+      purge_soft_delete_on_destroy = true
```
KeionneDerousselle, line 6: `purge_soft_delete_on_destroy` seems like the opposite of `recover_soft_deleted_key_vaults`. I'm not really sure why this combination of settings worked. I wasn't able to get it working (able to destroy and apply back to back) without both these settings set to true.

src/steps/resource-manager/monitor/constants.ts
```diff
@@ -28,4 +41,32 @@ export const MonitorRelationships = {
...
+  AZURE_RESOURCE_HAS_MONITOR_DIAGNOSTIC_LOG_SETTING: {
+    _type: 'azure_resource_has_diagnostic_log_setting',
+    sourceType: ANY_RESOURCE,
+    _class: RelationshipClass.HAS,
+    targetType: MonitorEntities.DIAGNOSTIC_LOG_SETTING._type,
+  },
+
+  AZURE_RESOURCE_HAS_MONITOR_DIAGNOSTIC_METRIC_SETTING: {
+    _type: 'azure_resource_has_diagnostic_metric_setting',
+    sourceType: ANY_RESOURCE,
```
KeionneDerousselle, line 54: What's the different between `ANY_RESOURCE` and `ANY_SCOPE`, @ndowmon?

src/steps/resource-manager/monitor/client.ts
```diff
@@ -32,4 +56,39 @@ export class MonitorClient extends Client {
...
+    return iterateAllResources({
+      logger: this.logger,
+      serviceClient,
+      resourceEndpoint: {
+        list: async () =>
+          serviceClient.diagnosticSettings
+            .list(resourceUri)
+            .then(
+              (result) =>
+                new DiagnosticSettingsListResponse(
```
KeionneDerousselle, line 84: The return type of the endpoint was not castable to type `ResourceListResponse`, so I had to create a return type that was.

src/steps/resource-manager/monitor/constants.ts
ndowmon: `ANY_SCOPE` ought to include non-resource scopes such as `subscription` or `resourceGroup`.

src/steps/resource-manager/monitor/diagnostic-settings.ts
```diff
@@ -0,0 +1,165 @@
...
+// TODO: Add the other types of storage when we ingest them. (Azure Log Analytics, Azure Event Hub)
+const STORAGE_LOCATION_INFO: StorageLocation[] = [
+  {
+    getStorageId: (diagnosticSettings: DiagnosticSettingsResource) =>
+      diagnosticSettings.storageAccountId,
+    storageType: STORAGE_ACCOUNT_ENTITY_METADATA._type,
+  },
+];
+
+async function createEntitiesAndRelationships(
```
ndowmon, line 69: I would probably make this function more explicit about what it does.

src/steps/resource-manager/monitor/diagnostic-settings.ts
```diff
@@ -0,0 +1,165 @@
...
+};
+
+// TODO: Add the other types of storage when we ingest them. (Azure Log Analytics, Azure Event Hub)
+const STORAGE_LOCATION_INFO: StorageLocation[] = [
+  {
+    getStorageId: (diagnosticSettings: DiagnosticSettingsResource) =>
+      diagnosticSettings.storageAccountId,
+    storageType: STORAGE_ACCOUNT_ENTITY_METADATA._type,
+  },
+];
```
ndowmon, line 67: We should chat about a good way to set up some of these maps. Also wondering if all of the interfaces are necessary.

src/steps/resource-manager/monitor/converters.ts
```diff
@@ -31,3 +36,83 @@ export function createMonitorLogProfileEntity(
...
+
+  return createSettingEntity(
+    webLinker,
+    diagnosticSetting,
+    setting,
+    MonitorEntities.DIAGNOSTIC_METRIC_SETTING._type,
+    MonitorEntities.DIAGNOSTIC_METRIC_SETTING._class,
+    id,
+  );
+}
```
ndowmon, line 118: These converters have me a little confused. It seems like there is one `DiagnosticSettingsResource` that contains an array of `MetricSettings` and `LogSettings`. 

If that's true, shouldn't there just be a `azure_diagnostic_settings` entity with all of these properties? And Then an `azure_diagnostic_settings_metrics` entity with the information about the `MetricSettings` and an `azure_diagnostic_settings_logs` entity with information about the `LogSettings`?

I would also expect the relationships to storage account to relate to `azure_diagnostic_settings`, not the `metrics` or `logs`.

ndowmon (changes requested):
Hey @KeionneDerousselle this is an awesome first pass. I didn't even fully realize the complexity of these entities and relationships.

I have added a ton of feedback to this PR. It would be great to find some time to catch up tomorrow about this! Hit me up to schedule some time once you've read this over.

src/steps/resource-manager/monitor/constants.ts
KeionneDerousselle: I think we should maybe use `ANY_SCOPE` then. I think you can have Diagnostic Settings on subscriptions (i.e. replacing Log Profiles)

terraform/keyvault.tf
```diff
@@ -35,3 +35,28 @@ resource "azurerm_key_vault_access_policy" "j1dev" {
     "get",
   ]
 }
+
+resource "azurerm_storage_account" "j1dev_key_vault_diag_set_strg" {
+  # The name can only consist of lowercase letters and numbers, and must be between 3 and 24 characters long.
+  name                     = "j1devkeyvaultdiagsetstrg"
```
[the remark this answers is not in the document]
KeionneDerousselle: The unique identifier is a good point. Also, no we don't need to create a new resource group or storage account. I usually do to be able to have separate unaffected environments that can be stood up on their own, but in this case it costs money...sooo....I'll use the existing resources :)

terraform/monitor.tf
```diff
@@ -56,3 +63,67 @@ resource "azurerm_monitor_log_profile" "j1dev_log_profile" {
...
+  }
+
+  metric {
+    category = "AllMetrics"
+
+    retention_policy {
+      enabled = false
+    }
+  }
+}
```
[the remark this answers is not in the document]
KeionneDerousselle: I created both of those so that I can have Key Vault tests, and tests that test the generic log to get diagnostic settings. I wasn't sure how to unit test it without the current recording setup we have now.

src/steps/resource-manager/monitor/diagnostic-settings.ts
```diff
@@ -0,0 +1,165 @@
+import {
```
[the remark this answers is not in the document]
KeionneDerousselle: Thanks. I will make that change.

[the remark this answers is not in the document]
KeionneDerousselle: I understand what you're saying. I put it in monitor because it uses the monitor client and monitor constants.

src/steps/resource-manager/monitor/diagnostic-settings.ts
```diff
@@ -0,0 +1,165 @@
...
+        fromType: settingEntity._type,
+        fromKey: settingEntity._key,
+        toType: storageType,
+        toKey: storageId,
+      }),
+    );
+  }
+}
+
+export async function fetchDiagnosticSettings(
```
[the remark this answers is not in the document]
KeionneDerousselle: I agree. That's a better name, and I will make that change.

src/steps/resource-manager/monitor/diagnostic-settings.ts
```diff
@@ -0,0 +1,165 @@
...
+  executionContext: IntegrationStepContext,
+  resourceEntity: Entity,
+): Promise<void> {
+  const { instance, logger, jobState } = executionContext;
+  const accountEntity = await jobState.getData<Entity>(ACCOUNT_ENTITY_TYPE);
+  const webLinker = createAzureWebLinker(accountEntity.defaultDomain as string);
+  const client = new MonitorClient(instance.config, logger);
+
+  const { id } = resourceEntity;
+  if (id && typeof id === 'string') {
```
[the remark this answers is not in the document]
KeionneDerousselle: I don't know that, in reality, it's ever undefined or an array, but typescript was complaining that it could be. I also figured we didn't want it to error out, but instead try to collect the other entities and relationships that it could.

src/steps/resource-manager/monitor/__recordings__/iterateDiagnosticSettings_3260250107/recording.har
```diff
@@ -0,0 +1,508 @@
...
+          "headersSize": 1677,
+          "httpVersion": "HTTP/1.1",
+          "method": "GET",
+          "queryString": [
+            {
+              "name": "api-version",
+              "value": "2016-06-01"
+            }
+          ],
+          "url": "https://management.azure.com/subscriptions?api-version=2016-06-01"
```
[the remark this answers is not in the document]
KeionneDerousselle: I'm not really sure. I don't think I specifically called that....is that something the endpoint does?

src/steps/resource-manager/monitor/client.ts
```diff
@@ -32,4 +56,39 @@ export class MonitorClient extends Client {
...
+                  result._response,
+                  result.value || [],
+                ),
+            ),
+      },
```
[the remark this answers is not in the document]
KeionneDerousselle: Ok

src/steps/resource-manager/monitor/converters.ts
ndowmon: You don't have to worry about this comment. It's alright to create these two entities as you and @aiwilliams have discussed.

src/steps/resource-manager/monitor/diagnostic-settings.ts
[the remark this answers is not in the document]
ndowmon:
```typescript
(id as string)
```

src/steps/resource-manager/monitor/diagnostic-settings.ts
```diff
@@ -0,0 +1,165 @@
...
+        _class: RelationshipClass.USES,
+        fromType: settingEntity._type,
+        fromKey: settingEntity._key,
+        toType: storageType,
+        toKey: storageId,
+      }),
+    );
+  }
+}
+
```
ndowmon, line 120:
```typescript
let monitorClient: MonitorClient;

function getMonitorClient(config, logger) {
  if (!monitorClient) { 
    monitorClient = new MonitorClient(config, logger);
  } 
 return monitorClient;
}

// inside fetchDiagnosticSettings()...

  // const client = new MonitorClient(instance.config, logger);
  const client = getMonitorClient(instance.config, logger);

```

src/steps/resource-manager/monitor/constants.ts
ndowmon: Can you please add docstrings on `ANY_SCOPE` and `ANY_RESOURCE` to describe the differences?

terraform/monitor.tf
```diff
@@ -97,7 +81,7 @@ resource "azurerm_monitor_diagnostic_setting" "j1dev_diag_set" {
   count              = local.monitor_diagnostic_settings_count
   name               = "j1dev_diag_set"
   target_resource_id = azurerm_key_vault.j1dev_diag_set_key_vault[0].id
```
ndowmon, line 83: Can you just use the existing `azurerm_key_vault.j1dev` resource? If you check out `keyvault.tf` you'll see that it's an always-on resource (it does not have a `create_key_vault_resources` flag like the `create_monitor_diagnostic_settings` flag we see in this file).

src/utils/diagnosticSettings.ts
```diff
@@ -1,124 +1,103 @@
 import {
```
ndowmon, line 1: Nice refactor! Let's still look through it in a little more detail together but this looks better to me. One change I think would be nice is to change the filename from `diagnosticSettings.ts` to match the primary (only?) export from this file.

terraform/monitor.tf
```diff
@@ -14,7 +20,7 @@ resource "azurerm_resource_group" "j1dev_log_profile_resource_group" {
 }
 
 resource "azurerm_storage_account" "j1dev_log_profile_storage_account" {
-  count = local.monitor_log_profile_count
+  count                    = local.monitor_log_profile_count
   # The name can only consist of lowercase letters and numbers, and must be between 3 and 24 characters long.
   name                     = "j1devlogprofilestrgacct"
```
ndowmon, line 25: Shoot this also needs to use the `developerId` TF variable since it the `name` of a storage account must be globally unique. Sorry I missed this in the **Log Profile** PR. Can you update?

src/steps/resource-manager/key-vault/index.ts
```diff
@@ -74,6 +86,10 @@ export const keyvaultSteps: Step<
       createResourceGroupResourceRelationshipMetadata(
         KEY_VAULT_SERVICE_ENTITY_TYPE,
       ),
+      MonitorRelationships.AZURE_RESOURCE_HAS_MONITOR_DIAGNOSTIC_LOG_SETTING,
+      MonitorRelationships.AZURE_RESOURCE_HAS_MONITOR_DIAGNOSTIC_METRIC_SETTING,
+      MonitorRelationships.DIAGNOSTIC_LOG_SETTING_USES_STORAGE_ACCOUNT,
+      MonitorRelationships.DIAGNOSTIC_METRIC_SETTING_USES_STORAGE_ACCOUNT,
```
ndowmon, line 92: Taking another look at this... it would be great if these could be exported as `diagnosticSettingsRelationshipsForResource` (an array of relationship metadata) from the `createDiagnosticSettingsEntitiesAndRelationshipsForResource` file. These same four relationships will show up _every_ time `createDiagnosticSettingsEntitiesAndRelationshipsForResource` is called in a step.

The same goes for `diagnosticSettingsEntitiesForResource`, which would be an array of entity metadata including `MonitorEntities.DIAGNOSTIC_LOG_SETTING` and `MonitorEntities.DIAGNOSTIC_METRIC_SETTING`.

src/steps/resource-manager/monitor/client.ts
```diff
@@ -32,4 +35,38 @@ export class MonitorClient extends Client {
...
+
+    return iterateAllResources({
+      logger: this.logger,
+      serviceClient,
+      resourceEndpoint: {
+        list: async () => {
+          const response = await serviceClient.diagnosticSettings.list(
+            resourceUri,
+          );
+          const diagnosticSettings = response.value!;
```
ndowmon, line 62: What happens if you try to call this endpoint on the following:
1. The resourceUri for a resource that exists and _can_ have a diagnostic setting, but does not (such as a key vault that has not provisioned diagnostic settings)
1. The resourceUri for a resource that exists but can _not_ have diagnostic settings. (You could pick a resource you've deployed through terraform that does not match the list that Azure publishes: https://docs.microsoft.com/en-us/azure/azure-monitor/platform/resource-logs-categories)
1. The resourceUri for a resource that does _not_ exist. E.g. `/subscriptions/40474ebe-55a2-4071-8fa8-b610acdd8e56/resourcegroups/j1dev_diag_set_resource_group/providers/microsoft.keyvault/vaults/J1_NON_EXISTENT_KEYVAULT`

I'm particularly concerned about the result of point 1, because we will be iterating every key vault after this PR is merged and we need to make sure the API response for key vaults without these configured diagnostic settings do not cause system errors.

ndowmon (commented):
Hey @KeionneDerousselle thanks for the revisions. I did another review of the changes and noticed a few things I missed previously.

src/steps/resource-manager/monitor/client.ts
ndowmon: Per slack, Point 1 returns an empty array.

ndowmon (approved):
Thank you so much for the iterations @KeionneDerousselle!

austinkelleher (approved):

PR #204, opened 2020-12-18
Upgrade SDK devDeps and increment SDK peerDep range

austinkelleher:


aiwilliams:
@austinkelleher @ndowmon seems this can be closed at this point eh?

PR #206, opened 2020-12-29
Feature/diagnostic settings activity logs

KeionneDerousselle:
# Acceptance Criteria
    
- 5.1.1 Ensure that a 'Diagnostics Setting' exists (Automated)
    - This CIS Benchmark aims to ensure that Diagnostic Settings are enabled for exporting Activity Logs.
    
# What Changed?
- Collected the **Diagnostic Settings** entities and relationships for an Azure **Subscription** (also known as **Actvity Logs**)
    - **Azure Diagnostic Log Setting** entity
    - **Azure Diagnostic Metric Setting** entity
    - **Azure Resource** has **Azure Diagnostic Log Setting** relationship
    - **Azure Resource** has **Azure Diagnostic Metric Setting** relationship
    - **Azure Diagnostic Log Setting** uses **Azure Storage Account** relationship
    - **Azure Diagnostic Metric Setting** uses **Azure Storage Account** relationship
    
# Graph
Screen Shot 2020-12-29 at 2 40 27 PM

### DIFF
--- a/src/steps/resource-manager/monitor/converters.ts
+++ b/src/steps/resource-manager/monitor/converters.ts
@@ -55,15 +55,16 @@ function createSettingEntity(
   } = diagnosticSetting;
 
   const { retentionPolicy, enabled, category, timeGrain } = setting;
+  const normalizedId = normalizeId(id);
 
   return createIntegrationEntity({
     entityData: {
       source: diagnosticSetting,
       assign: {
-        _key: id,
+        _key: normalizedId,
         _type,
         _class,
-        webLink: webLinker.portalResourceUrl(diagnosticSetting.id),
+        webLink: webLinker.portalResourceUrl(normalizeId(diagnosticSetting.id)),
         storageAccountId,
         serviceBusRuleId,
         eventHubAuthorizationRuleId,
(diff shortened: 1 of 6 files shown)


src/steps/resource-manager/monitor/converters.ts
```diff
@@ -75,12 +76,22 @@ function createSettingEntity(
...
+        id: normalizedId,
       },
     },
   });
 }
 
+/**
+ * When getting Diagnostic Settings for some Azure Resources, the Diagnostic Settings id is returned from the client without the leading slash.
+ * This function adds the leading slash to an id, if it does not exist
+ * @param id The resource URI of an Azure resource
```
aiwilliams, line 88: Gotta love the value we add by making data consistent!

src/steps/resource-manager/subscriptions/index.test.ts
```diff
@@ -1,58 +1,144 @@
...
+      expect.objectContaining({
+        _class: MonitorEntities.DIAGNOSTIC_LOG_SETTING._class,
+        _key: `/subscriptions/${instanceConfig.subscriptionId}/providers/microsoft.insights/diagnosticSettings/j1dev_sub_diag_set/logs/Policy/true/undefined/undefined`,
+        _type: MonitorEntities.DIAGNOSTIC_LOG_SETTING._type,
+        category: 'Policy',
+        displayName: 'j1dev_sub_diag_set',
+        enabled: true,
+        eventHubAuthorizationRuleId: undefined,
+        eventHubName: undefined,
+        id: `/subscriptions/${instanceConfig.subscriptionId}/providers/microsoft.insights/diagnosticSettings/j1dev_sub_diag_set/logs/Policy/true/undefined/undefined`,
```
aiwilliams, line 80: What risks would there be in changing the key generator for the diagnostic setting entity to avoid `undefined/undefined`? I'm working on memory usage problems in another integration, so I'm a bit sensitive ATM 😆  Perhaps we could put lighter-weight placeholders in these positions, so that we maintain a stable order, such as `0/0` when they're both `undefined`.

aiwilliams (approved):
Looks good, though I have a question, which can be addressed in a different PR since it will impact all diagnostic setting entity key/id (and relationships to them) generation.

src/steps/resource-manager/subscriptions/index.test.ts
aiwilliams: Thinking about this more with @KeionneDerousselle and @austinkelleher, seems we shouldn't care about optimizing every key, but focus on optimizing the things that keep references to the keys (the DuplicateKeyTracker) by hashing large keys. Anyway, nothing to do right now :)

PR #207, opened 2021-01-12
Fix how RM exposes ECONNRESET and node-fetch errors

ndowmon:
The RM client was not properly handling system / `node-fetch` errors, which caused errors to be exposed like:

```
Error: Provider API failed at storage.tables: undefined undefined
```

The proper error message should have been something like 
```
Error: Provider API failed at storage.tables: ECONNRESET Connection reset by server
```
or something similar.

More PRs to come, including retries and re-using this logic in Azure AD (Graph) APIs.

### DIFF
--- a/src/azure/resource-manager/client.ts
+++ b/src/azure/resource-manager/client.ts
@@ -5,12 +5,14 @@ import {
   RequestPolicyFactory,
   systemErrorRetryPolicy,
   throttlingRetryPolicy,
+  RestError as AzureRestError,
 } from '@azure/ms-rest-js';
 import {
   IntegrationLogger,
   IntegrationProviderAPIError,
 } from '@jupiterone/integration-sdk-core';
 import { retry } from '@lifeomic/attempt';
+import { FetchError } from 'node-fetch';
 
 import { IntegrationConfig } from '../../types';
 import authenticate from './authenticate';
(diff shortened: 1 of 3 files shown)


src/azure/resource-manager/client.ts
```diff
@@ -229,6 +231,53 @@ export interface IterateAllResourcesOptions<ServiceClientType, ResourceType> {
...
+      if (err instanceof AzureRestError) {
+        throw new IntegrationProviderAPIError({
+          cause: err,
+          endpoint: resourceDescription,
+          status: err.body?.code as string,
+          statusText: err.body?.message as string,
+        });
+      }
+      if (err instanceof FetchError) {
+        throw new IntegrationProviderAPIError({
```
austinkelleher, line 264: Do these types of errors get bubbled up to us? If a provider API is consistently failing (e.g. 500), it would be great for us to become aware of this even if it's not a bug on our side.

austinkelleher (approved):
Approved with question

src/azure/resource-manager/client.ts
aiwilliams: Yes, I believe these will be reported to us in Sentry.

ndowmon: Yes, `IntegrationProviderAPIError` reaches us and also pages in sentry.

PR #209, opened 2021-01-13
Retry non-Azure RM API errors

ndowmon:
Add mechanism to Azure Resource Manager client to retry non-Azure `RestError` errors, including `RestError`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Fixed
+
+- Started retrying all request errors that are not handled by the Azure API
+  client. Continue retrying 429 errors thrown by the Azure API client.
+
 ## 5.10.0 - 2020-01-13
 
 ### Added
--- a/src/azure/resource-manager/client.ts
+++ b/src/azure/resource-manager/client.ts
@@ -140,6 +140,7 @@ export abstract class Client {
 function retryResourceRequest<ResponseType>(
   requestFunc: () => Promise<ResponseType>,
   endpointRatePeriod: number,
+  logger: IntegrationLogger,
 ): Promise<ResponseType> {
   return retry(
     async (_context) => {
@@ -161,9 +162,25 @@ function retryResourceRequest<ResponseType>(
       // Most errors will be handled by the request policies. They will raise
       // a `RestError.statusCode: 429` when they see two 429 responses in a row,
       // which is the scenario we're aiming to address with our retry.
+      //
+      // Non Azure `RestError`s, such as ECONNRESET, should be retried.
       handleError: (err, context, _options) => {
-        if (err.statusCode !== 429) {
+        if (err instanceof AzureRestError && err.statusCode !== 429) {
+          logger.info(
+            {
+              err,
+            },
+            'Encountered non-retryable error in Resource Manager client.',
+          );
           context.abort();
+        } else {
+          logger.info(
+            {
+              err,
+              attemptsRemaining: context.attemptsRemaining,
+            },
+            'Encountered retryable error in Resource Manager client.',
+          );
         }
       },
     },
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -18,6 +18,10 @@ and this project adheres to
 
 - Fixed the way `IntegrationProviderAPIError` exposes error code/message of
   `node-fetch` errors, such as `ECONNRESET`.
+- Started retrying non-Azure error codes for Resource Manager API. The Azure
+  service client handles all non-429 responses with its own retry logic, but we
+  need to explicitly retry system errors such as ECONNRESET along with 429
+  errors.
```
aiwilliams, line 24: This confused me for a bit as I sought to understand how it is that the code will retry "ECONNRESET _along with 429 errors_". I came to see that the code we maintain still will not itself retry a `429` response. 👍 

How about this instead to simplify/clarify:
```suggestion
- Retry all request errors that are not handled by the Azure API client
```

ndowmon: @aiwilliams 
> I came to see that the code we maintain still will not itself retry a 429 response.

As you see below, our code will retry a 429 thrown from Azure itself. That was the initial use of this `retryResourceRequest` function, whose functionality I have expanded here.

```typescript
      // Most errors will be handled by the request policies. They will raise
      // a `RestError.statusCode: 429` when they see two 429 responses in a row,
      // which is the scenario we're aiming to address with our retry.
      //
      // Non Azure `RestError`s, such as ECONNRESET, should be retried.
      handleError: (err, context, _options) => {
        if (err instanceof AzureRestError && err.statusCode !== 429) {
          logger.info(
            {
              err,
            },
            'Encountered non-retryable error in Resource Manager client.',
          );
          context.abort();
        } else {
          logger.info(
            {
              err,
              attemptsRemaining: context.attemptsRemaining,
            },
            'Encountered retryable error in Resource Manager client.',
          );
```

src/azure/resource-manager/client.ts
```diff
@@ -161,9 +162,25 @@ function retryResourceRequest<ResponseType>(
       // Most errors will be handled by the request policies. They will raise
       // a `RestError.statusCode: 429` when they see two 429 responses in a row,
       // which is the scenario we're aiming to address with our retry.
+      //
+      // Non Azure `RestError`s, such as ECONNRESET, should be retried.
       handleError: (err, context, _options) => {
-        if (err.statusCode !== 429) {
+        if (err instanceof AzureRestError && err.statusCode !== 429) {
```
mknoedel, line 168: This seems good to me! I do wonder if we want to nail down further what errors we should be retrying. For instance, do we want to retry 500 errors?

CHANGELOG.md
mknoedel: Perhaps: "Start retrying all request errors that are not handled by the Azure API client. Continue retrying 429 errors thrown by the Azure API client."?

mknoedel (dismissed):
Some comments but looks great to me!

src/azure/resource-manager/client.ts
ndowmon: @mknoedel according to a comment added by @aiwilliams, those 500s should already be handled by the azure service clients. It appears that 429s are special cases:

```      
      // Most errors will be handled by the request policies. They will raise
      // a `RestError.statusCode: 429` when they see two 429 responses in a row,
      // which is the scenario we're aiming to address with our retry.
```

This will also begin logging any errors that are handled here and retried... we can use that information in the future to further hone this retry logic, if necessary.

CHANGELOG.md
aiwilliams: Thanks @ndowmon, looks like I was wr... wro... less right than I thought. 😁  Makes me think I ought not to work too late into the evening.

aiwilliams (dismissed):

CHANGELOG.md
ndowmon: Updated the language

aiwilliams (approved):

PR #210, opened 2021-01-13
Retry azure graph API calls

ndowmon:
Add a `request` function to the Azure Graph API, which centralizes requests and error handling for `GraphRequest.get()` requests, used to fetch Azure Active Directory resources.

Also add a `retryRequest` function to retry requests for Azure Graph API.

### DIFF
--- a/src/azure/graph/client.test.ts
+++ b/src/azure/graph/client.test.ts
@@ -2,10 +2,12 @@ import {
   createMockIntegrationLogger,
   Recording,
 } from '@jupiterone/integration-sdk-testing';
+import { FetchError } from 'node-fetch';
 
 import config from '../../../test/integrationInstanceConfig';
 import { GraphClient } from './client';
 import { setupAzureRecording } from '../../../test/helpers/recording';
+import { GraphRequest } from '@microsoft/microsoft-graph-client';
 
 class AnyGraphClient extends GraphClient {}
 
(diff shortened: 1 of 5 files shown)


mknoedel (approved):
LGTM. Idk if it is worth making a test or two testing the retry and error handling logic or not.

src/azure/graph/client.ts
```diff
@@ -27,13 +38,75 @@ export abstract class GraphClient {
...
+          endpoint,
+          status: err.statusCode,
+          statusText: err.statusText,
+        });
+      }
+
+      this.logger.warn(
+        { err, resourceUrl: endpoint },
+        'Encountered non-fatal error in Azure Graph client. Continuing step execution.',
+      );
```
aiwilliams, line 108: I think it would be good to avoid knowledge of the caller, both in terms of separation of concerns but also because this could be wrong at some point since it doesn't directly control step lifecycle.

aiwilliams (approved):

src/azure/graph/client.test.ts
```diff
@@ -55,3 +57,43 @@ test('fetchOrganization', async () => {
...
+  } as unknown) as GraphRequest;
+
+  await expect(client.request(mockGraphRequest)).rejects.toThrow(
+    'Provider API failed at https://hostname/endpoint: 400 Server Error',
+  );
+
+  expect(mockGet).toHaveBeenCalledTimes(3);
+});
+
+test.only('client.request should expose node-fetch error codes', async () => {
```
aiwilliams, line 81: `test.only`?

src/azure/graph/client.test.ts
```diff
@@ -55,3 +57,43 @@ test('fetchOrganization', async () => {
     ],
   });
 });
+
+test.only('client.request should retry requests 3 times', async () => {
```
aiwilliams, line 61: `test.only`?

ndowmon: Good catch!

Fixed. Thanks.

src/azure/graph/client.ts
```diff
@@ -27,13 +38,75 @@ export abstract class GraphClient {
...
+    return this.request(this.client.api('/'));
   }
 
   public async fetchOrganization(): Promise<Organization> {
-    const response = await this.client.api('/organization').get();
-    return response.value[0];
+    const response = await this.request<Organization>(
+      this.client.api('/organization'),
+    );
+    return response!.value[0];
```
aiwilliams, line 49: I guess it's best to allow this to kill the program when we don't get the organization data back? I see it basically used to work this way.

aiwilliams (approved):

src/azure/graph/client.ts
ndowmon: Yes, I kept the functionality as it was for now. If this becomes an issue in the future we can always change it.

PR #217, opened 2021-02-02
Diagnostic Settings for Azure Event Grid

KeionneDerousselle:
# Acceptance Criteria

-  5.3 Ensure that Diagnostic Logs are enabled for all services which support it. (Automated)
    - Unlike diagnostic logs that capture activity for the subscription level (Activity Log), there are also logs for actions taken on specific resources. This benchmark aims to ensure that resource-level diagnostic logs provide insight into operations that were performed within that resource itself. 

    - There are several Azure resources that support diagnostic logging. Event Grid Domains and Event Grid Topics are two of those Azure resources.
   
# What Changed?
- Collected the **Diagnostic Settings** entities and relationships for Azure **Event Grid Domains** and Azure **Event Grid Topics**
    - **Azure Diagnostic Log Setting** entity
    - **Azure Diagnostic Metric Setting** entity
    - **Azure Resource** has **Azure Diagnostic Log Setting** relationship
    - **Azure Resource** has **Azure Diagnostic Metric Setting** relationship
    - **Azure Diagnostic Log Setting** uses **Azure Storage Account** relationship
    - **Azure Diagnostic Metric Setting** uses **Azure Storage Account** relationship
    
# Graph

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -1,5 +1,5 @@
 provider "azurerm" {
-  version = "~>2.29.0"
+  version = "~>2.43.0"
   features {
     key_vault {
       recover_soft_deleted_key_vaults = true
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Added
+
+- Diagnostic Settings entities and relationships for Activity Logs (Azure
+  Subscription)
+
 ## 5.10.1 - 2020-01-15
 
 ### Fixed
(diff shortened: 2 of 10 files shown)


terraform/main.tf
```diff
@@ -1,5 +1,5 @@
 provider "azurerm" {
-  version = "~>2.29.0"
+  version = "~>2.43.0"
```
ndowmon, line 2: Interesting... did this version have to do with the issues we were seeing in `azurerm_role_definition` provisioning?

ndowmon (approved):
Looks great!

PR #227, opened 2021-02-26
Add match-requests-by for azure

ndowmon:


src/azure/resource-manager/authenticate.ts
```diff
@@ -5,14 +5,52 @@ import { IntegrationConfig } from '../../types';
...
+  const response = await loginWithServicePrincipalSecretWithAuthResponse(
+    config,
+  );
+
+  if (
+    !response.subscriptions ||
+    !response.subscriptions.find((s) => s.id === config.subscriptionId)
+  ) {
+    throw new IntegrationValidationError(
+      'subscriptionId not found in tenant specified by directoryId',
```
ndowmon, line 23: We only need to do this check for initial validation. Not for every auth call. This caused issues when the polly matcher began working even when the subscriptionId changed, which is why I pulled it out into an explicit validation function.

austinkelleher (approved):

PR #228, opened 2021-02-26
Cis benchmark 1.3 iam

ndowmon:


test/helpers/__recordings__/matchRequestsBy-graph_277829644/recording.har
```diff
@@ -226,18 +226,23 @@
...
...1f7-4c69-a1f2-f497d3f11ced\",\"userType\":\"Member\"},{\"businessPhones\":[],\"displayName\":\"Austin Kelleher\",\"givenName\":null,\"jobTitle\":null,\"mail\":\"austin.kellher@jupiterone.com\",\"mobilePhone\":null,\"officeLocation\":null,\"preferredLanguage\":null,\"surname\":null,\"userPrincipalName\":\"austin.kellher_jupiterone.com#EXT#@ndowmongmail.onmicrosoft.com\",\"id\":\"d87d02c5-e07e-4093-a274-0232864312fe\",\"userType\":\"Guest\"},{\"businessPhones\":[],\"displayName\":\"Keionne Derousselle\",\"givenName\":null,\"jobTitle\":null,\"mail\":\"kanddventurepartners@gmail.com\",\"mobilePhone\":null,\"officeLocation\":null,\"preferredLanguage\":null,\"surname\":null,\"userPrincipalName\":\"kanddventurepartners_gmail.com#EXT#@ndowmongmail.onmicrosoft.com\",\"id\":\"621154ca-e35e-4ac8-80a1-f02f495ce68a\",\"userType\":\"Guest\"},{\"businessPhones\":[],\"displayName\":\"Nick Dowmon\",\"givenName\":\"Nick\",\"jobTitle\":null,\"mail\":null,\"mobilePhone\":null,\"officeLocation\":null,\"preferredLanguage\":\"en\",\"surname\":\"Dowmon\",\"userPrincipalName\":\"ndowmon_gmail.com#EXT#@ndowmongmail.onmicrosoft.com\",\"id\":\"b6e7f627-8731-47bd-be0e-477d1cbc6e17\",\"userType\":\"Member\"}]}"
```
ndowmon, line 245: If you open this file in a good diff viewer (e.g. vscode), you'll see that this recording returns everything from the previous, _plus_ the `userType` property.

src/steps/active-directory/converters.ts
```diff
@@ -91,7 +91,7 @@ export function createGroupEntity(data: Group): Entity {
         name: data.displayName,
         deletedOn: getTime(data.deletedDateTime),
         createdOn: getTime(data.createdDateTime),
-        email: data.mail,
+        email: data.mail ?? undefined,
```
ndowmon, line 94: I added `toMatchGraphObjectSchema` for these AD steps, and the schema matching failed because `data.mail` can be `null`. Which does not validate our schema.

src/steps/active-directory/client.ts
```diff
@@ -100,18 +110,19 @@ export class DirectoryGraphClient extends GraphClient {
...
     callback: (item: T) => void | Promise<void>;
   }): Promise<void> {
     let nextLink: string | undefined;
     do {
       let api = this.client.api(nextLink || resourceUrl);
-      if (query) {
-        api = api.query(query);
+      api.select;
+      if (options?.select) {
+        api = api.select(options.select);
```
ndowmon, line 125: I was testing this on the `user` endpoint and noticed that the microsoft graph client exposes a `select` method, which seems to be the encouraged API from microsoft.

src/steps/active-directory/index.test.ts
```diff
@@ -11,279 +14,188 @@ import {
...
+    }
+  }
+  return { targets, rest };
+}
+
 test('active directory steps', async () => {
   recording = setupAzureRecording({
     directory: __dirname,
     name: 'active-directory-steps',
+    options: { matchRequestsBy: getMatchRequestsBy({ config: configFromEnv }) },
```
ndowmon, line 125: These tests required a re-record, because the query string changed, so I modernized them.

aiwilliams (dismissed):

src/steps/active-directory/index.test.ts
```diff
@@ -11,279 +14,188 @@ import {
...
+  expect(groupEntities).toMatchGraphObjectSchema({
+    _class: GROUP_ENTITY_CLASS,
+  });
+
+  // expect(groupMemberEntities.length).toBeGreaterThan(0); // The 'ad-group-members' can simply return relationships in some cases.
+  expect(groupMemberEntities).toMatchGraphObjectSchema({
+    _class: GROUP_MEMBER_ENTITY_CLASS,
+  });
+
+  expect(restEntities.length).toBe(0);
```
mknoedel, line 163: Is this important? This makes it so this step isn't expandable in the future without changing this test

src/steps/active-directory/client.ts
```diff
@@ -83,7 +75,25 @@ export class DirectoryGraphClient extends GraphClient {
...
+      'jobTitle',
+      'mail',
+      'mobilePhone',
+      'officeLocation',
+      'preferredLanguage',
+      'surname',
+      'userPrincipalName',
+      'id',
+    ];
+    const select = [...defaultSelect, 'userType'];
```
mknoedel, line 91: Is this going to need to be updated in the MS365 integration as well?

mknoedel (approved):

src/steps/active-directory/client.ts
ndowmon: It probably should be. In general, I think at some point we're going to want to pull out the active directory steps into their own integration, and have this project and `microsoft-365` import `steps` and `validateInvocation` from that project.

src/steps/active-directory/index.test.ts
ndowmon: That's the idea. Should this test add additional entities, it will ensure that this test accounts for them.

mknoedel: My philosophy with tests is that if engineers need to update your test to add something, that is an indication that the test is overreaching. What you are saying with this check is that no one should ever add more entities or relationships to the active directory steps which I am not so sure is what you are trying to say.

src/steps/active-directory/client.ts
mknoedel: That makes a ton of sense. Probably should have done that straight out. I guess for now I'll make a quick story to update the AD steps of MS365 while keeping in mind that this refactor will come at some point.

src/steps/active-directory/index.test.ts
ndowmon: I am saying that. I consider this test suite to be integration testing rather than unit testing.

My message is "this is the comprehensive test for active directory steps. If you add something to the active directory steps, and you don't update these comprehensive tests, they're no longer comprehensive, which is their intention."

ndowmon: I am saying that. I consider this test suite to be integration testing rather than unit testing.

My message is "this is the comprehensive test for active directory steps. If you add something to the active directory steps, and you don't update these comprehensive tests, they're no longer comprehensive, which is their intention."

PR #233, opened 2021-03-09
Add blob service properties to azure_storage_account

ndowmon:


src/steps/resource-manager/storage/converters.ts
```diff
@@ -51,9 +53,10 @@ function getArrayOfStorageAccountEndpoints(
 export function createStorageAccountEntity(
   webLinker: AzureWebLinker,
   data: StorageAccount,
+  storageBlobServiceProperties?: BlobServiceProperties,
```
ndowmon, line 56: Wondering if this ought to be required, not optional.

src/steps/resource-manager/storage/index.test.ts
```diff
@@ -18,190 +22,159 @@ afterEach(async () => {
...
 
-test('step - storage accounts', async () => {
-  const instanceConfig: IntegrationConfig = {
-    clientId: process.env.CLIENT_ID || 'clientId',
-    clientSecret: process.env.CLIENT_SECRET || 'clientSecret',
-    directoryId: 'bcd90474-9b62-4040-9d7b-8af257b1427d',
-    subscriptionId: '40474ebe-55a2-4071-8fa8-b610acdd8e56',
-  };
+describe('step - storage accounts', () => {
+  async function getSetupEntities() {
```
ndowmon, line 26: I changed up the testing pattern here in order to actually fetch the `accountEntity` and `keyVaultEntity` required as setup for this test. I an interested to hear feedback on this pattern.

terraform/storage.tf
```diff
@@ -36,6 +36,12 @@ resource "azurerm_storage_account" "j1dev" {
   }
 }
 
+resource "azurerm_storage_account_customer_managed_key" "j1dev" {
```
ndowmon, line 39: There was not actually a link between `azure_storage_account|USES|azure_key_vault` in the terraform, so I needed to add it in order to make my tests pass.

src/steps/resource-manager/storage/converters.ts
ndowmon: 👍

test/helpers/recording.ts
```diff
@@ -80,7 +80,33 @@ export function getMatchRequestsBy({ config }: { config: IntegrationConfig }) {
     headers: false,
     url: {
       pathname: (pathname) => {
-        return pathname.replace(config.directoryId, 'directory-id');
+        pathname = pathname.replace(config.directoryId, 'directory-id');
+        if (!(pathname as string).startsWith('//subscriptions')) {
```
ndowmon, line 84: bonus 😇

package.json
```diff
@@ -20,10 +20,10 @@
     "lint:es": "eslint . --cache --fix --ext .ts,.tsx",
     "format": "prettier --write '**/*.{ts,js,json,css,md,yml}'",
     "type-check": "tsc",
-    "test": "jest",
-    "test:ci": "yarn lint && yarn type-check && yarn test",
+    "test": "LOAD_ENV=1 jest",
+    "test:ci": "yarn lint && yarn type-check && yarn jest",
     "prebuild": "yarn test:ci",
-    "prepush": "yarn lint && yarn type-check && yarn test",
+    "prepush": "yarn test:ci",
```
ndowmon, line 26: I got burned by failed CI builds because this prepush was testing with my loaded `.env`. This'll avoid that in the future - thanks for the pattern in GCP @austinkelleher

test/integrationInstanceConfig.ts
```diff
@@ -1,4 +1,12 @@
 import { IntegrationConfig } from '../src/types';
+import * as dotenv from 'dotenv';
+import * as path from 'path';
+
+if (process.env.LOAD_ENV) {
```
mknoedel, line 5: 👍

test/helpers/recording.ts
```diff
@@ -80,7 +80,33 @@ export function getMatchRequestsBy({ config }: { config: IntegrationConfig }) {
...
+          // instance config (and should be replaced in the recording).
+          //
+          // ``` typescript
+          // const resourceGroupName = 'j1dev';
+          // const keyVaultName = 'ndowmon11-j1dev';
+          // const path = `https://management.azure.com/subscriptions/${config.subscriptionId}/resourceGroups/${resourceGroupName}/providers/Microsoft.KeyVault/vaults/${keyVaultName}`;
+          // const response = await fetch(path);
+          // ``
+          pathname = pathname.replace(
+            config.subscriptionId || 'subscription-id',
```
mknoedel, line 105: why `|| 'subscription-id`?

test/helpers/recording.test.ts
```diff
@@ -6,11 +6,16 @@ import { J1SubscriptionClient } from '../../src/steps/resource-manager/subscript
 import {
```
mknoedel, line 6: Having a recording.test is a good idea, I'll start adding them to projects I'm working on going forward.

mknoedel (approved):
Looks great!

test/helpers/recording.ts
mknoedel: I'm having a hard time figuring out where is the `//subscriptions` coming from. Is that just the pattern that Azure uses in their URL generation?

ndowmon: So, given the example documented below:

```typescript
const resourceId = '/subscriptions/<s-id>/resourceGroups/<rg-id>/providers/Microsoft.KeyVault/vaults/<kv-id>';
const path = `https://management.azure.com/${resourceId}/providers/microsoft.insights/diagnosticSettings`;
```

The constructed URL will have a double forward slash because of how it was constructed:

```typescript
const path = 'https://management.azure.com//subscriptions/<s-id>/resourceGroups/<rg-id>/providers/Microsoft.KeyVault/vaults/<kv-id>/providers/microsoft.insights/diagnosticSettings';
```

It's visible in the recording here: https://github.com/JupiterOne/graph-azure/blob/cis-benchmark-3.8/src/steps/resource-manager/storage/__recordings__/resource-manager-step-storage-accounts_690738595/recording.har#L1508


I'm definitely partial to use `matchRequestsBy` in our graph projects over hard-coding the credentials used to record a specific API call. This particular matcher may not work for all cases, but we can certainly continue to iterate if new cases show up.

package.json
ndowmon: Our integration template definitely needs a refresh...

test/helpers/recording.ts
ndowmon: `config.subscriptionId` is an optional argument. Turns out this wasn't strictly necessary, as passing `undefined` as the first variable to `String.replace()` doesn't appear to have side-effects:

```bash
nickdowmon@Nicks-MacBook-Pro graph-azure % node
Welcome to Node.js v14.9.0.
Type ".help" for more information.
> 'some-string'.replace(undefined, 'replacement')
'some-string'
```

test/helpers/recording.test.ts
ndowmon: I found this particularly useful because of the `matchRequestsBy` function. If we move in that direction, I would like to see it tested any of our projects moving forward.

test/helpers/recording.ts
mknoedel: Oh, I see now thanks!

mknoedel (approved):

PR #242, opened 2021-03-19
Cis benchmark 6.5

ndowmon:
subscription-location

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to
 
 ### Added
 
+- Added `azure_location` entities.
 - Added `azure_network_watcher` entities.
 - Added `azure_security_group_flow_logs` entities.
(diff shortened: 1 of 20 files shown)


mknoedel:
Did you configure all of those locations or are those always pulled for every subscription? If those are always pulled maybe there is a way to see what locations are actually used or something

mknoedel (approved):
Looks great!  Have a comment above

ndowmon:
> Did you configure all of those locations or are those always pulled for every subscription? If those are always pulled maybe there is a way to see what locations are actually used or something

@mknoedel that's a good question. As far as I can tell, these are static for the subscriptions, but I suppose the enumerations grow as Azure adds new regions.

In any case, seems that this enumeration will be necessary to answer Azure CIS Benchmark 6.5

mknoedel (approved):

PR #243, opened 2021-03-21
Cis benchmark 7.1

ndowmon:
The `usesManagedDisk` property will help to answer CIS Benchmark 7.1.

I've done some testing of Azure VMs on my own, and it **appears** that if the VM uses a managed OS disk, it _must_ use managed data disks, and if the VM uses an unmanaged OS disk, it _must_ use unmanaged data disks. If this were true, we could simply check if the OS disk is managed to answer this question.

In any case, the implementation of `usesManagedDisks` is thorough enough to handle even cases where this isn't true.

Additionally, this change fixes a bug I noticed when creating VM -> disk relationships.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to
 
 ### Added
 
+- Added `azure_vm.usesManagedDisk` property.
 - Added `azure_location` entities.
 - Added `azure_network_watcher` entities.
 - Added `azure_security_group_flow_logs` entities.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to
 ### Added
 
 - Added `azure_managed_disk.encryption` property.
-- Added `azure_vm.usesManagedDisk` property.
+- Added `azure_vm.usesManagedDisks` property.
 - Added `azure_location` entities.
 - Added `azure_network_watcher` entities.
 - Added `azure_security_group_flow_logs` entities.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,13 @@ and this project adheres to
 - Added `azure_security_group_flow_logs` entities.
 - Added `azure_sql_server_firewall_rule` entities.
 
+### Fixed
+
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -10,11 +10,20 @@ and this project adheres to
 
 ### Added
 
+- Added `azure_managed_disk.encryption` property.
+- Added `azure_vm.usesManagedDisk` property.
```
ndowmon, line 14: This really should be `usesManagedDisks`, to be consistent with both Azure & CIS documentation... changing.

src/steps/resource-manager/compute/index.ts
```diff
@@ -54,12 +57,89 @@ export async function fetchVirtualMachines(
...
+      return createDirectRelationship({
+        _class: RelationshipClass.USES,
+        from: vmEntity,
+        to: managedDiskEntity,
+        properties: {
+          [diskType]: true,
+        },
+      });
+    } else {
+      logger.error(
```
mknoedel, line 99: If I were @aiwilliams I would say that logger.error should likely be reserved for debugging errors so this might want to be logger.warn instead.

mknoedel (approved):

src/steps/resource-manager/compute/index.ts
ndowmon: I personally think this would be a debug-worthy error - if `vm.storageProfile.osDisk.managedDisk.id` is not undefined, but we can't find the entity in our job state, something has gone wrong with how I've implemented this and I'd want to be alerted.

aiwilliams: I do think that if we do not terminate the step, then we should not log an `error` because we just handled the unexpected behavior and carried on with ingestion. If the program is truly broken in its implementation, we should throw an `IntegrationError` and get alerted to fix the integration, which will provide the stack trace and be logged as an error.

Level 50 as warnings is really confusing during incident response. It is very helpful during IR to be able to query for level 50 messages and find that they are related to an incident.

ndowmon: Thanks for the comment @aiwilliams. I have changed this to `logger.warn` in #255

PR #244, opened 2021-03-21
Add azure_managed_disk.encryption property

ndowmon:


src/steps/resource-manager/compute/converters.ts
```diff
@@ -98,6 +98,7 @@ export function createDiskEntity(
     createdOn: getTime(data.timeCreated),
     webLink: webLinker.portalResourceUrl(data.id),
     encrypted: !!data.encryption?.type,
+    encryption: data.encryption?.type,
```
austinkelleher, line 101: The AWS integration uses `encryptionAlgorithm` for several entities. Example query in prod:

```
/*
Returns:

aws_s3_bucket
aws_dynamodb_table
*/
find unique * with encryptionAlgorithm!=undefined as e return e._type
```

austinkelleher (changes requested):
I think we should consider changing this property name to `encryptionAlgorithm`

src/steps/resource-manager/compute/converters.ts
ndowmon: The value of this is something like `EncryptionAtRestWithCustomerKey`, not an encryption algorithm.

austinkelleher (approved):

PR #247, opened 2021-03-23
Upgrade sdk 5.11.0

ndowmon:
`getData<T>(key: string)` now returns `Promise<T|undefined>`, so I did the simple fix for all existing instances of this function.

### DIFF
--- a/src/steps/active-directory/index.ts
+++ b/src/steps/active-directory/index.ts
@@ -66,7 +66,7 @@ export async function fetchUsers(
   const { logger, instance, jobState } = executionContext;
   const graphClient = new DirectoryGraphClient(logger, instance.config);
 
-  const accountEntity = await jobState.getData<Entity>(ACCOUNT_ENTITY_TYPE);
+  const accountEntity = (await jobState.getData<Entity>(ACCOUNT_ENTITY_TYPE))!;
   await graphClient.iterateUsers(async (user) => {
     const userEntity = createUserEntity(user);
     await jobState.addEntity(userEntity);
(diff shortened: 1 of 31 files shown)


src/steps/active-directory/index.ts
```diff
@@ -66,7 +66,7 @@ export async function fetchUsers(
   const { logger, instance, jobState } = executionContext;
   const graphClient = new DirectoryGraphClient(logger, instance.config);
 
-  const accountEntity = await jobState.getData<Entity>(ACCOUNT_ENTITY_TYPE);
+  const accountEntity = (await jobState.getData<Entity>(ACCOUNT_ENTITY_TYPE))!;
```
austinkelleher, line 69: This is minor, but I created a helper function in the Jamf integration that throws an error if this entity is not found instead of using the `!` operator everywhere. It will ensure that if we forget to add the `dependsOn`, that we will exit in a meaningful way instead of trying to build the relationship. This can also be caught in tests assuming we have a test for every case of this.

austinkelleher (approved):
I approved with a comment

src/steps/active-directory/index.ts
austinkelleher: This one: https://github.com/JupiterOne/graph-jamf/blob/master/src/util/account.ts#L15

mknoedel (approved):

src/steps/active-directory/index.ts
ndowmon: Good design. Thanks for sharing.

austinkelleher (approved):

PR #248, opened 2021-03-23
Cis benchmark 6.5

ndowmon:
In order to answer CIS benchmark 6.5, we will have to be able to relate the network watchers with locations inside of a subscription

### DIFF
--- a/src/steps/resource-manager/network/index.test.ts
+++ b/src/steps/resource-manager/network/index.test.ts
@@ -10,6 +10,7 @@ import {
 import {
   getMatchRequestsBy,
   setupAzureRecording,
+  shouldReplaceSubscriptionId,
 } from '../../../../test/helpers/recording';
 import { fetchAccount } from '../../active-directory';
 import {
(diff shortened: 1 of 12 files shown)


src/steps/resource-manager/network/index.test.ts
```diff
@@ -2118,3 +2125,169 @@ describe('rm-network-flow-logs', () => {
...
+                  !isListLocationsEndpoint(pathname)
+                ) {
+                  pathname = pathname.replace(
+                    configFromEnv.subscriptionId || 'subscription-id',
+                    'subscription-id',
+                  );
+                }
+                return pathname;
+              },
+            },
```
mknoedel, line 2208: Wow, this got complicated :P

mknoedel (approved):

src/steps/resource-manager/network/index.test.ts
ndowmon: Yeah... this _mostly_ duplicates what's in `getMatchRequestsBy`, but adds one logical condition. I didn't want to update the `getMatchRequestsBy` because reasons so I duplicated it here.

PR #251, opened 2021-03-23
Cis benchmark 7.7

ndowmon:


src/steps/resource-manager/compute/index.ts
```diff
@@ -58,23 +63,62 @@ export async function fetchVirtualMachines(
...
+      const storageAccountBlobEndpoint = storageAccount?.primaryEndpoints?.blob;
+
+      if (!storageAccountBlobEndpoint) {
+        logger.warn(
+          {
+            storageAccountId: storageAccount?.id,
+            storageAccountBlobEndpoint,
+          },
+          'Could not find blob endpoint for storage account; not adding storage account to blobEndpointStorageAccountMap',
+        );
```
aiwilliams, line 89: 💯  I'm thinking `warn` is good when we skip something so that when a customer asks about their integration missing something, we can search for level 40 messages.

src/steps/resource-manager/compute/index.ts
```diff
@@ -111,6 +155,39 @@ export async function createVirtualMachineDiskRelationships(
     }
   }
 
+  function createUnmanagedDiskRelationship(options: {
```
aiwilliams, line 158: The `createVirtualMachineDiskRelationships` function has gotten quite long with lots of conditional logic. These "inner functions" are helpful for organizing the code. Thanks for the tests for all this!

aiwilliams (dismissed):

src/steps/resource-manager/compute/index.ts
ndowmon: 🙂

ndowmon:
@aiwilliams I've merged everything down. I addressed your [comment](https://github.com/JupiterOne/graph-azure/pull/253#discussion_r601432535) as well - can you verify & approve?

aiwilliams (approved):

PR #258, opened 2021-03-25
Cis benchmark 4.5

ndowmon:
The tests needed to be re-recorded because of an additional API call in the `sql_server` step, which caused a lot of the test setup to fail. I refactored these steps, including breaking out a separate diagnostic settings step.

The diagnostic settings need to be re-constructred, which will happen in the next PR.


src/steps/resource-manager/databases/sql/constants.ts
```diff
@@ -19,6 +21,9 @@ export const entities = {
     _type: 'azure_sql_server',
     _class: RM_DATABASE_SERVER_ENTITY_CLASS,
     resourceName: '[RM] SQL Server',
+    rawDataKeys: {
+      ENCRYPTION_PROTECTOR: 'encryptionProtector',
+    },
```
ndowmon, line 26: I'd like to start utilizing a set of `rawDataKeys` to set / access raw data on entities.

mknoedel: is there a reason why we wouldn't just pull those values into the entity instead?

src/steps/resource-manager/databases/sql/index.ts
```diff
@@ -49,20 +52,53 @@ export async function fetchSQLDatabases(
...
+  const webLinker = createAzureWebLinker(accountEntity.defaultDomain as string);
+  const client = new SQLClient(instance.config, logger);
+
+  await jobState.iterateEntities(
+    { _type: entities.SERVER._type },
+    async (serverEntity) => {
+      const server = getRawData<Server>(serverEntity, 'default');
+      if (!server) return;
 
-    try {
```
mknoedel, line 65: I believe removing this `try` `catch` makes it so one api fail stops all later database entity creations? Seems like we might be able to implement a pattern where we aggregate errors and then throw one at the end so we can get most of the data. Idk if that is the right thinking though. My thought is most data is better than no data.

mknoedel (approved):

src/steps/resource-manager/databases/sql/constants.ts
ndowmon: We can put those values on the entity, but i think it is safer to use the types provided by the API providers (Azure) when referencing values in downstream steps. It's a pattern I'd like to test out in this project and see how it goes.

```typescript
function stepOne(context) {
  const entity = createIntegrationEntity({ _type: 'my_type', ...properties});
  const encryptionProtector = await getEncryptionProtector(context);
  entity['encryptionProtector.serverKeyName'] = encryptionProtector.serverKeyName;
  setRawData(entity, { name: 'encryptionProtector', rawData: encryptionProtector});
}

function stepTwo(context) {
  await context.jobState.iterateEntities({ _type: 'my_type'}, async (entity) => {
    const encryptionProtector = getRawData<EncryptionProtector>(entity, 'encryptionProtector');

    const serverKeyName = encryptionProtector?.serverKeyName;
    // encryption protector is typed, so it's safer to use than a property on the entity.
    // const serverKeyName = entity['encryptionProtector.serverKeyName'];
    await createSomeSpecialRelationship(context, entity, serverKeyName);
  });
}
```

src/steps/resource-manager/databases/sql/index.ts
ndowmon: That's true. This try/catch initially implemented primarily because this step was trying to do more than one thing - get the servers, and then get the databases. 

I think the appropriate place for a try/catch would be here (https://github.com/JupiterOne/graph-azure/blob/master/src/azure/resource-manager/client.ts#L351), where every step could leverage this logic. However, it also will require a way to mark the step's dataset as `partial`.

I've opened an issue to track here: https://github.com/JupiterOne/graph-azure/issues/260

aiwilliams: > so we can get most of the data. Idk if that is the right thinking though. My thought is most data is better than no data.

I'm fully convinced that's the right thinking, and the `partial` mechanism is meant to support this, though we certainly have room to improve! I've shared [additional thoughts on that new Issue](https://github.com/JupiterOne/graph-azure/issues/260#issuecomment-808370430), thanks @ndowmon !

src/steps/resource-manager/databases/sql/constants.ts
aiwilliams: The AWS integration uses the pattern you're talking about @ndowmon in various ways. I think it does have it's place. Also, I would like to avoid creating specific sub-types of `Entity` just to "safely" reference properties on the entities later, so I'm a fan. 😄 

Do note that when we have a way to declare relationships between things in a "mini mapper", it will be necessary to add properties to entities "just to make the mapping work".

PR #270, opened 2021-03-30
Add azure_monitor_activity_log_alert entities

ndowmon:
Screen Shot 2021-03-30 at 1 33 03 PM


austinkelleher (approved):

src/steps/resource-manager/monitor/constants.ts
```diff
@@ -18,6 +30,30 @@ export const MonitorEntities = {
     _class: ['Configuration'],
     resourceName: '[RM] Monitor Diagnostic Settings Resource',
   },
+  ACTIVITY_LOG_ALERT: {
+    _type: 'azure_monitor_activity_log_alert',
+    _class: ['Rule'],
+    resourceName: '[RM] Monitor Activity Log Alert',
+    schema: {
```
mknoedel, line 37: What's this? Are we supposed to be adding schemas to entities?

ndowmon: More experimentation. Added a proposal to the SDK here: https://github.com/JupiterOne/sdk/issues/449

ndowmon: For the record, I'm using this schema in `toMatchGraphObjectSchema({ _class, schema })`, and I would like to generate docs from this spec.

src/steps/resource-manager/monitor/index.test.ts
```diff
@@ -77,3 +94,159 @@ test('step = monitor log profiles', async () => {
...
+      options: {
+        matchRequestsBy: getMatchRequestsBy({ config: configFromEnv }),
+      },
+    });
+
+    const {
+      accountEntity,
+      subscriptionEntity,
+      activityLogAlertEntities,
+    } = await getSetupEntities(configFromEnv);
```
mknoedel, line 234: I stole this pattern in a recent Intune PR and I think it should just be the default pattern. Makes things much cleaner once you know what's going on (which I didn't at first.)

src/steps/resource-manager/monitor/index.ts
```diff
@@ -62,6 +73,86 @@ export async function fetchLogProfiles(
...
+
+export async function buildActivityLogScopeRelationships(
+  executionContext: IntegrationStepContext,
+): Promise<void> {
+  const { logger, jobState } = executionContext;
+  await jobState.iterateEntities(
+    { _type: MonitorEntities.ACTIVITY_LOG_ALERT._type },
+    async (alertEntity) => {
+      const alert = getRawData<ActivityLogAlertResource>(alertEntity);
+      if (!alert) return;
```
mknoedel, line 128: 1. Would this actually be a bug if there is no alert
2. Are we worried that the alert doesn't have scopes? Because if it doesn't it will throw a really weird type error.

mknoedel (approved):
Looks great!

src/steps/resource-manager/monitor/index.test.ts
ndowmon: Nice! Maybe in a coming sprint we can spend some time moving these patterns back into https://github.com/JupiterOne/integration-template.

src/steps/resource-manager/monitor/index.ts
ndowmon: 1. Yes, I think this probably _would_ be a bug. Although I am really hoping it doesn't happen. Added this issue to add later: https://github.com/JupiterOne/graph-azure/issues/272
2. The types provided by azure specify that `scopes` is required, so I'm comfortable not checking this.

PR #285, opened 2021-04-06
Only ingest one subscription

mknoedel:
Now that we are making tools for clients to make an Azure integration instance per Subscription, it no longer makes sense to ingest all subscriptions in every integration run. 

After talking with @ndowmon, the solution is to only ingest subscription data based on the passed in `config.subscriptionId`


src/steps/resource-manager/subscriptions/client.ts
```diff
@@ -46,4 +48,38 @@ export class J1SubscriptionClient extends Client {
       callback,
     });
   }
+
+  public async fetchSubscriptions(): Promise<Array<Subscription> | undefined> {
```
mknoedel, line 52: This method makes the code I am using in the `jupiter-integration-azure` project more clean. It is not used in this graph project though.

CHANGELOG.md
```diff
@@ -8,6 +8,11 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Changed
+
+- Only igest the subscription resource provided in the `subscriptionId` config
+  field.
```
aiwilliams, line 14: I think it helpful to explain these kinds of changes a bit further, to help folks understand why we made this change. How about:
```suggestion
- Only ingest an entity for the subscription identified by the `subscriptionId` config
  field. Previously, the integration would create an entity for every subscription, though
  none of the resources would be ingested. This new behavior limits the scope to the
  subscription defined by `subscriptionId`.
```

aiwilliams (approved):

src/steps/resource-manager/network/index.test.ts
```diff
@@ -1414,7 +1414,7 @@ describe('rm-network-location-watcher-relationships', () => {
           },
         },
       });
-    });
+    }, 10000);
```
ndowmon, line 1417: Not a big deal, but after recording, I prefer to remove the `timeout` arg from these tests.

src/steps/resource-manager/subscriptions/client.ts
```diff
@@ -3,8 +3,10 @@ import { Subscription, Location } from '@azure/arm-subscriptions/esm/models';
 import {
   Client,
   iterateAllResources,
+  request,
 } from '../../../azure/resource-manager/client';
 
+const FIVE_MINUTES = 5 * 60 * 1000;
 export class J1SubscriptionClient extends Client {
   public async iterateSubscriptions(
```
ndowmon, line 11: Does this method get used anywhere in the project anymore? If not, it may be worth removing this dead code and any related tests.

src/steps/resource-manager/subscriptions/index.ts
```diff
@@ -31,7 +32,16 @@ export async function fetchSubscriptions(
   const webLinker = createAzureWebLinker(accountEntity.defaultDomain as string);
```
ndowmon, line 32: Maybe the name of this function should be changed from `fetchSubscriptions` to `fetchSubscription`.

src/steps/resource-manager/subscriptions/index.ts
```diff
@@ -31,7 +32,16 @@ export async function fetchSubscriptions(
...
+  if (!instance.config.subscriptionId) {
+    // This should never happen as getStepStartStates should turn off this step if there is no subscriptionId
+    throw new IntegrationConfigLoadError(
+      'You need to provide a subscriptionId in order to ingest a subscription',
+    );
+  }
+  const subscription = await client.fetchSubscription(
+    instance.config.subscriptionId,
+  );
+  if (subscription) {
```
ndowmon, line 44: If `subscription=undefined`, I think it'd be best to throw an exception here - it will affect a ton of downstream steps and we'll need all those steps to be marked as `partial`.

ndowmon (approved):
Great change! Thanks for taking this one. I've added a few comments.

src/steps/resource-manager/network/index.test.ts
aiwilliams: Wouldn't re-recording then require added this back again? Perhaps the timeout should be set at the top of the file?

mknoedel: This test failed due to a timeout so I just added this timeout here to move on. idk if we want a more permeant solution or not.

src/steps/resource-manager/subscriptions/client.ts
mknoedel: I left it in in case we wanted to add a feature to ingest the subscriptions like is being done today. If you think that won't matter I'll go ahead and delete it.

src/steps/resource-manager/subscriptions/index.ts
mknoedel: Sounds like a good idea to me thanks!

src/steps/resource-manager/subscriptions/index.ts
mknoedel: makes sense. I'll make all these methods singular.

src/steps/resource-manager/subscriptions/client.ts
```diff
@@ -57,7 +62,14 @@ export class J1SubscriptionClient extends Client {
       },
     );
     const subscriptions = await request(
-      async () => await serviceClient.subscriptions.list(),
+      // Need API version to 2020-01-01 in order to return subscription tags
```
ndowmon, line 65: 😳  bummer. 

I think switching to this package (https://www.npmjs.com/package/@azure/arm-subscriptions-profile-2020-09-01-hybrid) might have worked, but not sure it's a better solution necessarily.

ndowmon (approved):

PR #290, opened 2021-04-08
Decouple subscription & diagnostic setting entities

ndowmon:
Because my Azure subscription has an associated diagnostic setting, the tests made an additional API call and the matchers didn't work properly. This primarily decouples the `rm-subscription` step from retrieving diagnostic settings, a pattern we've moved towards in other steps in this project.

It also renames a bit of the testing infrastructure for the subscription step.


src/steps/resource-manager/subscriptions/index.test.ts
```diff
@@ -96,6 +91,10 @@ describe('step - subscription', () => {
   });
 });
 
+describe.skip('rm-subscription-diagnostic-settings', () => {
+  // Not implementing tests for diagnostic settings at this time.
```
ndowmon, line 95: Steps that _only_ fetch diagnostic settings don't offer much value in testing this.

src/steps/resource-manager/subscriptions/index.ts
```diff
@@ -62,6 +57,22 @@ export async function fetchSubscription(
   }
 }
 
+export async function fetchSubscriptionDiagnosticSettings(
```
ndowmon, line 60: These execution handlers can be further abstracted, but maybe i'll save that for another day.

aiwilliams (approved):

PR #301, opened 2021-04-21
Add try/catch logic to AzurePolicyClient methods for policy definitions

ndowmon:
```
### Changed

- Added try/catch logic to `AzurePolicyClient` methods for policy definitions &
  policy set definitions so step does not fail from 404s.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Changed
+
+- Added try/catch logic to `AzurePolicyClient` methods for policy definitions &
+  policy set definitions so step does not fail from 404s.
+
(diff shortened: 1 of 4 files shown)


austinkelleher (commented):
I don't know all of the details of this change, so I'm speculating a bit here...

If any of these API calls fail for valid reasons (e.g. 500 errors), we should probably be marking the entire step as partial instead of just logging a warn, right? Otherwise, we will cause this data to be deleted when synchronized.

austinkelleher (approved):
Approved with question.

ndowmon:
Mentioned offline that this is primarily for visibility, will hopefully revert the try/catch once I find out how these failing requests ought to be appropriately handled

PR #303, opened 2021-04-22
Changelog for 5.23.4

ndowmon:
```
## 5.23.4 - 2021-04-22

### Changed

- Upgraded `@jupiterone/integration-sdk-*@6.0.0`
- Added try/catch logic to `AzurePolicyClient` methods for policy definitions &
  policy set definitions so step does not fail from 404s.
```


aiwilliams (approved):
Is this meant to change `package.json` and tag for release?

PR #307, opened 2021-04-27
Handle policyDefinitionId sourced from management groups

ndowmon:
```
### Added

- Added logic to handle `policyDefinition` IDs that are sourced from management
  groups.
```

During the original implementation, it wasn't clear that there were management group-based policy definitions, and these were mishandled causing some integrations to fail. Now that these are properly handled, I tore out a good amount of the safety netting and `logger.warn` statements so that this step _does_ fail when the API returns 4xx/5xx responses. See @austinkelleher 's comments in a previous PR https://github.com/JupiterOne/graph-azure/pull/301#pullrequestreview-641402702

### DIFF
--- a/src/steps/resource-manager/policy/client.ts
+++ b/src/steps/resource-manager/policy/client.ts
@@ -34,81 +34,67 @@ export class AzurePolicyClient extends Client {
 
   public async getPolicySetDefinition(
     name: string,
-  ): Promise<PolicySetDefinition | undefined> {
+  ): Promise<PolicySetDefinition> {
     const serviceClient = await this.getAuthenticatedServiceClient(
       PolicyClient,
     );
 
-    try {
-      return await serviceClient.policySetDefinitions.get(name);
-    } catch (err) {
-      this.logger.warn(
-        {
-          name,
-          err,
-        },
-        'Error getting policy set definition by name.',
-      );
-    }
+    return serviceClient.policySetDefinitions.get(name);
   }
 
   public async getBuiltInPolicySetDefinition(
     name: string,
-  ): Promise<PolicySetDefinition | undefined> {
+  ): Promise<PolicySetDefinition> {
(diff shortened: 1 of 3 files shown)


src/steps/resource-manager/policy/client.ts
```diff
@@ -34,81 +34,67 @@ export class AzurePolicyClient extends Client {
...
-    } catch (err) {
-      this.logger.warn(
-        {
-          name,
-          err,
-        },
-        'Error getting policy set definition by name.',
-      );
-    }
+    return serviceClient.policySetDefinitions.get(name);
```
mknoedel, line 42: Is this error getting handled elsewhere now?

src/steps/resource-manager/policy/findOrCreatePolicyDefinitionEntityFromAssignment.ts
```diff
@@ -108,16 +109,6 @@ async function findOrCreatePolicyDefinitionEntity(
         policyDefinitionId,
         name,
       );
-      if (!policyDefinition) {
```
mknoedel, line 111: Are we not concerned about not having a policyDefinition anymore?

mknoedel (approved):

src/steps/resource-manager/policy/client.ts
ndowmon: Nope. Originally there was no error handling, but I added it while investigating the source of this issue. Now I'm taking it back out, since all cases should be handled.

src/steps/resource-manager/policy/findOrCreatePolicyDefinitionEntityFromAssignment.ts
ndowmon: Nope - changed the interface to return `Promise<PolicyDefinition>` instead of `Promise<PoilcyDefinition | undefined>` so this check isn't necessary.

PR #320, opened 2021-05-01
Fix network firewall type value

aiwilliams:
@ndowmon this was on a list of mine to address after asking you a couple of months back whether `azure_network_azure_firewall` was the correct type and learning that it was an oversight.


austinkelleher (approved):
We'll have to ensure that we update any related managed questions

aiwilliams:
> We'll have to ensure that we update any related managed questions

Good call, looks like this one: https://github.com/JupiterOne/jupiter-integration-azure/blob/master/deploy/terraform/data/questions.yaml#L1269

mknoedel (approved):

ndowmon (approved):

Issue #321, opened 2021-05-03
Ingest private endpoint connections for azure_storage_account entities


ndowmon:

ndowmon:
Closed with #322

PR #322, opened 2021-05-04
Ingest private endpoints

ndowmon:
See #321 

Screen Shot 2021-05-04 at 3 52 08 PM

### DIFF
--- a/src/steps/resource-manager/network/index.test.ts
+++ b/src/steps/resource-manager/network/index.test.ts
@@ -22,6 +22,7 @@ import {
   fetchNetworkSecurityGroupFlowLogs,
   fetchNetworkSecurityGroups,
   fetchNetworkWatchers,
+  fetchPrivateEndpoints,
   fetchPublicIPAddresses,
   fetchVirtualNetworks,
 } from './';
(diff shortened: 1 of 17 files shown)


kenanwarren (dismissed):

src/steps/resource-manager/network/index.test.ts
```diff
@@ -1159,6 +1163,66 @@ describe('rm-network-watchers', () => {
...
+
+    const privateEndpointEntities = context.jobState.collectedEntities;
+
+    expect(privateEndpointEntities.length).toBeGreaterThan(0);
+    /**
+     * The Azure private endpoint does not comply with JupiterOne's network endpoint class, because
+     * J1's NetworkEndpoint requires the "ipAddress" property. However, the Azure private endpoint
+     * relates back to a network interface, which is where the IP address is defined.
+     */
+    // expect(privateEndpointEntities).toMatchGraphObjectSchema({
```
kenanwarren, line 1210: Should we update the data model? I'm not sure if there is a way to mark fields as either this or that in the data model. I feel like the private endpoint -> network interface for the ip isn't an uncommon thing.

ndowmon: Possibly. What would be the best move? Removing `ipAddress` from required fields on `NetworkEndpoint`?

kenanwarren: I think so, ideally we'd have a way to mark that either one field or another is required. I feel like the more integrations we get the less required fields will have.

austinkelleher (dismissed):

austinkelleher (approved):

PR #323, opened 2021-05-04
Ingest subnet -> private endpoint relationships

ndowmon:
See #321 

Note that this private endpoint resource has direct relationships to **resource group**, **subnet**, **network interface** and then a target resource (in this case, the storage account called `ndowmon11j1dev`).
Screen Shot 2021-05-04 at 4 50 33 PM

Screen Shot 2021-05-04 at 4 37 52 PM


kenanwarren (approved):
lgtm

PR #325, opened 2021-05-05
Add terraform for azurerm_private_endpoint

ndowmon:
Adds terraform for `azurerm_private_endpoint` and re-records a few tests.

### DIFF
--- a/terraform/compute.tf
+++ b/terraform/compute.tf
@@ -50,14 +50,16 @@ resource "azurerm_subnet" "j1dev" {
   name                 = "j1dev"
   resource_group_name  = azurerm_resource_group.j1dev.name
   virtual_network_name = azurerm_virtual_network.j1dev.name
-  address_prefix       = "10.0.2.0/24"
+  address_prefixes     = ["10.0.2.0/24"]
+  enforce_private_link_endpoint_network_policies = true
+  enforce_private_link_service_network_policies  = false
 }
 
 resource "azurerm_subnet" "j1dev_priv_one" {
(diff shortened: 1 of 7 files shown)


terraform/compute.tf
```diff
@@ -50,14 +50,16 @@ resource "azurerm_subnet" "j1dev" {
   name                 = "j1dev"
   resource_group_name  = azurerm_resource_group.j1dev.name
   virtual_network_name = azurerm_virtual_network.j1dev.name
-  address_prefix       = "10.0.2.0/24"
+  address_prefixes     = ["10.0.2.0/24"]
+  enforce_private_link_endpoint_network_policies = true
```
ndowmon, line 54: This value needs to be `true` so that we can create a private endpoint.

terraform/compute.tf
```diff
@@ -50,14 +50,16 @@ resource "azurerm_subnet" "j1dev" {
...
+  enforce_private_link_service_network_policies  = false
 }
 
 resource "azurerm_subnet" "j1dev_priv_one" {
   name                 = "j1dev_priv_one"
   resource_group_name  = azurerm_resource_group.j1dev.name
   virtual_network_name = azurerm_virtual_network.j1dev.name
-  address_prefix       = "10.0.3.0/24"
+  address_prefixes     = ["10.0.3.0/24"]
```
ndowmon, line 62: Added some bonus changes where the provider has deprecated features.

austinkelleher (approved):

kenanwarren (approved):

PR #329, opened 2021-05-07
Ingest app service apps

ndowmon:
The Azure endpoint for `Function App` resources (`providers/Microsoft.Web/sites`) is generic, and returns both function apps (`azure_function_app`) and web apps (`azure_web_app`). I chose to create these as two different `_type`s, because they're well segmented in the azure console, so normalizing on a single `_type` (like `azure_app`) seems super confusing.

That said, the best differentiator from the API response is the `type` property, which is just a comma-separated list that may include `functionapp` or `app`. Don't completely love that as a differentiator, so let me know if you think this should have been done differently.

Screen Shot 2021-05-07 at 12 20 54 PM

### DIFF
--- a/terraform/appservice.tf
+++ b/terraform/appservice.tf
@@ -2,8 +2,8 @@ resource "azurerm_app_service_plan" "j1dev" {
   name                = "ASP-j1dev-function-app"
   location            = azurerm_resource_group.j1dev.location
   resource_group_name = azurerm_resource_group.j1dev.name
-	kind                = "FunctionApp"
-	reserved            = true
+  kind                = "FunctionApp"
+  reserved            = true
 
   sku {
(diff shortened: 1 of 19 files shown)


terraform/appservice.tf
```diff
@@ -0,0 +1,44 @@
+resource "azurerm_app_service_plan" "j1dev" {
+  name                = "ASP-j1dev-function-app"
+  location            = azurerm_resource_group.j1dev.location
+  resource_group_name = azurerm_resource_group.j1dev.name
+	kind                = "FunctionApp"
```
kenanwarren, line 5: Formatting looks a little weird for these lines.

kenanwarren (approved):

PR #349, opened 2021-05-19
Ingest virtual machine vm/shared image relationships

### LINKED ISSUE #344: Ingest Managed Disk -> Image relationships
Requires https://github.com/JupiterOne/graph-azure/issues/343

Also might require ingestion of `Microsoft.Compute/snapshots`.

kenanwarren:
Resolves #344


src/steps/resource-manager/compute/index.ts
```diff
@@ -417,6 +417,36 @@ async function findOrCreateVirtualMachineExtensionEntity(
...
+  executionContext: IntegrationStepContext,
+): Promise<void> {
+  const { jobState } = executionContext;
+  const relationships: Relationship[] = [];
+
+  await jobState.iterateEntities(
+    { _type: VIRTUAL_MACHINE_ENTITY_TYPE },
+    async (vmEntity) => {
+      const vm = getRawData<VirtualMachine>(vmEntity);
+      const imageRefId = vm?.storageProfile?.imageReference?.id;
```
austinkelleher, line 430: In Google Cloud the relationships is:

```
Host -USES-> Disk -USES-> Image
```

I believe that the azure integration ingests VM disks. Should the relationship actually be through the disk? It would be great if we could create a shared query between these integrations.

```
find Host
THAT USES Disk
THAT USES Image
```

austinkelleher (approved):
Approved with a question regarding relationships

src/steps/resource-manager/compute/index.ts
ndowmon: @austinkelleher check out my recent comment on #344 https://github.com/JupiterOne/graph-azure/issues/344#issuecomment-844094035

kenanwarren: We talk a bit about it in #344, but the gist is a disk image in azure map to snapshots, but not the underlying images themselves.

Issue #341, opened 2021-05-17
Ingest Azure Continuous Delivery configurations


ndowmon:
<img width="856" alt="Screen Shot 2021-05-17 at 1 17 15 PM" src="https://user-images.githubusercontent.com/15333061/118530060-361b8680-b712-11eb-9039-77ba36e0dcc9.png">

kenanwarren:
Appears to just be an extension on the virtual machine:
```json
{
  "name": "TeamServicesAgentExtension",
  "id": "/subscriptions/zip/resourceGroups/zoop/providers/Microsoft.Compute/virtualMachines/zap/extensions/TeamServicesAgentExtension",
  "type": "Microsoft.Compute/virtualMachines/extensions",
  "location": "eastus",
  "properties": {
      "autoUpgradeMinorVersion": true,
      "provisioningState": "Succeeded",
      "publisher": "Microsoft.VisualStudio.Services",
      "type": "TeamServicesAgentLinux",
      "typeHandlerVersion": "1.0",
      "settings": {
          "VSTSAccountName": "zip",
          "TeamProject": "zoop",
          "DeploymentGroup": "test-deployment",
          "AgentName": "zap",
          "Tags": ""
      }
  }
}
```

kenanwarren:
With the open PR we'll ingest these settings to enable a j1ql query that could get this information. Ideally in the future we'd automatically link across integrations. (Azure -> Azure DevOps Integrations)

kenanwarren:
Closing with #350, to have the cross integration link we can make a new ticket.

PR #350, opened 2021-05-20
Add new vm extension properties to get things like CD settings

kenanwarren:
Partially solves #341 by allowing j1ql queries on extension settings where the continuous delivery information resides.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,11 @@ and this project adheres to
   | `azure_vm`             | `HAS`      | `azure_shared_image` |
   | `azure_vm`             | `HAS`      | `azure_image`        |
 
+- Changed the following property values on `azure_vm_extension`:
+  - `settings`
+  - `extType`
+  - `publisher`
+
 ## 5.24.0 - 2021-05-11
 
 ### Added
--- a/src/steps/resource-manager/compute/converters.ts
+++ b/src/steps/resource-manager/compute/converters.ts
@@ -137,7 +137,7 @@ export function createImageEntity(
 
 export type VirtualMachineExtensionSharedProperties = Omit<
   VirtualMachineExtension,
-  'id' | 'location' | 'provisioningState'
+  'location' | 'provisioningState'
 >;
 
 export function createVirtualMachineExtensionEntity(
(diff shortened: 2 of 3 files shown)


mknoedel (approved):

src/steps/resource-manager/compute/converters.ts
```diff
@@ -151,6 +151,9 @@ export function createVirtualMachineExtensionEntity(
         _type: entities.VIRTUAL_MACHINE_EXTENSION._type,
         _class: entities.VIRTUAL_MACHINE_EXTENSION._class,
         name: data.name,
+        publisher: data.publisher,
+        extType: data.virtualMachineExtensionType,
+        settings: data.settings,
```
aiwilliams, line 156: IIRC, `data.settings` is an Object. Are there any tests for this converter that use the schema-based validation test function? I hope it would expose this as a problem.

I was thinking the the `data.settings` would end up as:
```
{
  _type: entities.VIRTUAL_MACHINE_EXTENSION._type,
  ...,
  'settings.SomeThingDontKnow': data.settings['SomeThingDontKnow'],
```

Though it would be done using `convertProperties`, which can accomplish something like this I think.

The J1QL would end up being `find Application with settings.SomeThingDontKnow = 'expected value'`.

aiwilliams (changes requested):

src/steps/resource-manager/compute/converters.ts
kenanwarren: I don't believe this function has a test, I'll add one and switch over settings to use convert properties.

kenanwarren: Fixed!

CHANGELOG.md
```diff
@@ -26,6 +26,11 @@ and this project adheres to
   | `azure_vm`             | `HAS`      | `azure_shared_image` |
   | `azure_vm`             | `HAS`      | `azure_image`        |
 
+- Changed the following property values on `azure_vm_extension`:
```
aiwilliams, line 29: FYI, there is a desire to move to a format for new properties according to [a relatively new standard](https://github.com/JupiterOne/docs/pull/246/files#diff-ebe8b2e8480b4289d5f017a9a0415ad5cb2db37b8367f6a414591d8c87fb5d0aR118) for release notes.

aiwilliams (approved):
Looks good, thanks @kenanwarren!

PR #351, opened 2021-05-20
339 management groups

ndowmon:
This introduces a new config field `ingestManagementGroups` that should be used like the `ingestActiveDirectory` config field. However, it may be best to remove `intestManagementGroups` and just ingest them when `ingestActiveDirectory` is set to `true`.

Here's a look at the management group structure in Azure:
Screen Shot 2021-05-20 at 11 45 28 AM

Screen Shot 2021-05-20 at 11 44 52 AM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,10 +12,11 @@ and this project adheres to
 
 - Added support for ingesting the following **new** resources:
 
-  | Service       | Resource / Entity    |
-  | ------------- | -------------------- |
-  | Gallery       | `azure_gallery`      |
-  | Gallery Image | `azure_shared_image` |
+  | Service           | Resource / Entity        |
+  | ----------------- | ------------------------ |
+  | Gallery           | `azure_gallery`          |
+  | Gallery Image     | `azure_shared_image`     |
+  | Management Groups | `azure_management_group` |
 
 - Added support for ingesting the following **new** relationships:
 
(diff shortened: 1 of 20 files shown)


kenanwarren (dismissed):

aiwilliams:
> it may be best to remove `intestManagementGroups` and just ingest them when `ingestActiveDirectory` is set to `true`.

I think we should just ingest them when `ingestActiveDirectory: true`, but I'm curious to hear the reason you thought it something a user may not want to ingest.

aiwilliams:
Hmm, after further reflection and consideration of the code which references `@azure/arm-managementgroups` etc. (Resource Manager), I'm wondering if the management groups are dependent on ingesting AD at all. Why not just always ingest management groups?

kenanwarren:
I think we should ingest them irrespective of AD. If I recall you can "bucket" subscriptions into management groups and I can see that being pretty useful to writing queries that might be specific to one of those subscription buckets.

image

ndowmon:
>Hmm, after further reflection and consideration of the code which references @azure/arm-managementgroups etc. (Resource Manager), I'm wondering if the management groups are dependent on ingesting AD at all. Why not just always ingest management groups?

@aiwilliams and @kenanwarren my sentiment is that management groups should be ingested just _once_ for each directory. If we ingested management groups for every subscription, and a user ingested two or more subscriptions in the same directory, all of the management groups would be duplicated in J1.

ndowmon:
Actually, we should probably ingest management groups when **Configure Subscription Instances** is checked.

Screen Shot 2021-05-24 at 8 44 17 AM

aiwilliams:
> we should probably ingest management groups when **Configure Subscription Instances** is checked

That sounds great, thanks @ndowmon !

kenanwarren (dismissed):

kenanwarren (approved):

PR #352, opened 2021-05-20
Update build.yml

erichs-ext:


.github/workflows/build.yml
```diff
@@ -11,6 +11,12 @@ jobs:
         os: [ubuntu-latest]
 
     steps:
+      - name: NPM preflight
+        env:
+          NPM_AUTH_TOKEN: ${{ secrets.NPM_AUTH_TOKEN }}
+        run: |
+          echo "//registry.npmjs.org/:_authToken=${NPM_AUTH_TOKEN}" > .npmrc
```
mknoedel, line 18: Is this a test? Because it sure seems like we shouldn't merge this in.

kenanwarren:
Going forward please leave a detailed description of why you're making the requested change. This doesn't seem like something we'd like to add to our build pipeline.

PR #353, opened 2021-05-24
Add isMfaRegistered property to user

### LINKED ISSUE #346: Ingest User MFA Status
https://graph.microsoft.com/v1.0/users/{{user-id}}/authentication/microsoftAuthenticatorMethods

https://docs.microsoft.com/en-us/graph/api/microsoftauthenticatorauthenticationmethod-list?view=graph-rest-1.0&tabs=http

kenanwarren:
resolves #346

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,10 +26,12 @@ and this project adheres to
   | `azure_vm`             | `HAS`      | `azure_shared_image` |
   | `azure_vm`             | `HAS`      | `azure_image`        |
 
-- Changed the following property values on `azure_vm_extension`:
-  - `settings`
-  - `extType`
-  - `publisher`
+- New properties added to resources:
+
+  | Entity               | Properties                         |
+  | -------------------- | ---------------------------------- |
+  | `azure_vm_extension` | `settings`, `extType`, `publisher` |
+  | `azure_user`         | `isMfaRegistered`                  |
 
 ## 5.24.0 - 2021-05-11
 
--- a/src/steps/active-directory/converters.ts
+++ b/src/steps/active-directory/converters.ts
@@ -23,6 +23,7 @@ import {
   GroupMember,
   MemberType,
   IdentitySecurityDefaultsEnforcementPolicy,
+  CredentialUserRegistrationDetails,
 } from './client';
 import {
   ACCOUNT_ENTITY_CLASS,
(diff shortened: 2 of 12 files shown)


src/steps/active-directory/index.ts
```diff
@@ -83,8 +86,23 @@ export async function fetchUsers(
...
+    string,
+    CredentialUserRegistrationDetails
+  >();
+  await graphClient.iterateCredentialUserRegistrationDetails(
+    (credentialDetails) => {
+      if (credentialDetails.id) {
+        credentialDetailsMap.set(credentialDetails.id, credentialDetails);
+      }
+    },
+  );
```
ndowmon, line 99: I wonder if this logic should be placed in a separate step. You could possibly create this map and store it in the job state using `jobState.setData`, and retrieve it again in the user step. I'm just concerned that we don't accidentally cause the user step to fail in the event that this endpoint starts giving us trouble. What do you think?

src/steps/active-directory/converters.ts
```diff
@@ -127,6 +131,7 @@ export function createUserEntity(data: User): Entity {
         firstName: data.givenName,
         lastName: data.surname,
         username: data.userPrincipalName,
+        isMfaRegistered: credentialDetails?.isMfaRegistered,
       },
     },
   });
```
ndowmon, line 137: You can use `setRawData()` to also add the entirety of this `credentialDetails` object to the user entity. See https://github.com/JupiterOne/graph-azure/blob/master/src/steps/resource-manager/databases/converters.ts#L109

CHANGELOG.md
```diff
@@ -26,10 +26,12 @@ and this project adheres to
...
-- Changed the following property values on `azure_vm_extension`:
-  - `settings`
-  - `extType`
-  - `publisher`
+- New properties added to resources:
+
+  | Entity               | Properties                         |
+  | -------------------- | ---------------------------------- |
+  | `azure_vm_extension` | `settings`, `extType`, `publisher` |
+  | `azure_user`         | `isMfaRegistered`                  |
```
ndowmon, line 34: 💯 🥳

src/steps/active-directory/index.ts
kenanwarren: I think that would be fine. However, I do wrap the endpoint in a general try catch mainly because there are no guarantees it'll work given it's a beta endpoint and you need a specific class of azure active directory. Given that I'm not sure it can fail the user step right now.

src/steps/active-directory/converters.ts
kenanwarren: Yeah I wasn't sure if we wanted to ingest all the data.

src/steps/active-directory/index.ts
kenanwarren: updated!

ndowmon (approved):

CHANGELOG.md
aiwilliams: Yah, thanks for this formatting @kenanwarren, will make life easier next Monday! 🎉

PR #356, opened 2021-05-25
Decouple active directory steps

ndowmon:
We need to isolate the resource manager, active directory, and management group steps so that relationships are consistent no matter how the integration instance is configured.

```
### Changed

- Previously, JupiterOne created direct relationships between active directory
  entities and role assignments when active directory entities were ingested in
  the same integration as subscription entities. In an effort to ensure
  consistent relationships are built given any integration configuration, this
  integration now defaults to creating mapped relationships between role
  assignments and active directory entities.
```


mknoedel (commented):
Looks good to me, thought I'm not really sure of its implications. I'll hold off for a bit until others review it. Then I'll approve it.

src/getStepStartStates.test.ts
```diff
@@ -621,3 +626,103 @@ describe('getStepStartStates', () => {
...
+      );
+    }
+  });
+});
+
+declare global {
+  // eslint-disable-next-line @typescript-eslint/no-namespace
+  namespace jest {
+    interface Matchers<R> {
+      toHaveIsolatedDependencies(integrationSteps: Step<any>[]): R;
```
kenanwarren, line 699: Where do these come from?

kenanwarren (approved):

PR #366, opened 2021-05-31
Add azure vm -> managed identity relationships

ndowmon:
Azure Managed Identities are IAM identities assigned to compute relationships (such as VMs) that allow those compute resources to access other Azure resources using RBAC permissions. See https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/overview

Managed Identities can be either user-assigned or system-assigned. In both cases, managed identities themselves are nothing but an `azure_service_principal` with `"servicePrincipalType": "ManagedIdentity"` already ingested through Azure Active Directory ingestion.


src/steps/resource-manager/compute/index.test.ts
```diff
@@ -650,83 +660,238 @@ describe('rm-compute-virtual-machine-extensions', () => {
...
+  // eslint-disable-next-line @typescript-eslint/no-namespace
+  namespace jest {
+    interface Matchers<R> {
+      toCreateValidRelationshipsToEntities(entities: Entity[]): R;
+    }
+  }
+}
+
+expect.extend({
+  toCreateValidRelationshipsToEntities(
```
aiwilliams, line 841: Niice. Perhaps add an Issue to SDK project to extract something like this from here?

src/steps/resource-manager/compute/index.ts
```diff
@@ -447,6 +452,95 @@ export async function buildVirtualMachineImageRelationships(
...
+  const { jobState, logger } = executionContext;
+
+  await jobState.iterateEntities(
+    { _type: VIRTUAL_MACHINE_ENTITY_TYPE },
+    async (vmEntity) => {
+      const vm = getRawData<VirtualMachine>(vmEntity);
+
+      if (
+        vm?.identity?.type === 'SystemAssigned' ||
+        vm?.identity?.type === 'SystemAssigned, UserAssigned'
```
aiwilliams, line 467: Are these always going to be so specifically formed that using a regex isn't going to be necessary?

aiwilliams (approved):

src/steps/resource-manager/compute/index.test.ts
ndowmon: Yep. I've got a branch in the SDK with this, but want to write tests before opening PR.

src/steps/resource-manager/compute/index.ts
ndowmon: Yes - the azure SDK types this explicitly: https://github.com/Azure/azure-sdk-for-js/blob/master/sdk/compute/arm-compute/src/models/index.ts#L9464

```typescript
export type ResourceIdentityType = 'SystemAssigned' | 'UserAssigned' | 'SystemAssigned, UserAssigned' | 'None';
```

PR #391, opened 2021-06-07
387 vm shared image version relationships

### LINKED ISSUE #387: Relate virtual machines to shared versions instead of shared images
Technically, it's more correct to link virtual machines to shared image versions, rather than shared image definitions.

Prerequisite: #386

ndowmon:
Fixes #387 

```
- Added support for ingesting the following **new** relationships:

  | Source               | \_class | Target                       |
  | -------------------- | ------- | ---------------------------- |
  | `azure_vm`           | `USES`  | `azure_shared_image_version` |
```

Screen Shot 2021-06-06 at 10 57 41 PM

### DIFF
--- a/src/steps/resource-manager/compute/index.ts
+++ b/src/steps/resource-manager/compute/index.ts
@@ -475,7 +475,8 @@ export async function buildVirtualMachineImageRelationships(
     { _type: VIRTUAL_MACHINE_ENTITY_TYPE },
     async (vmEntity) => {
       const vm = getRawData<VirtualMachine>(vmEntity);
-      const imageRefId = vm?.storageProfile?.imageReference?.id;
+      const imageReference = vm?.storageProfile?.imageReference;
+      const imageRefId = imageReference?.id;
       if (!imageRefId) {
         return;
       }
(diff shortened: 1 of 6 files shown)


src/steps/resource-manager/compute/index.ts
```diff
@@ -489,6 +490,34 @@ export async function buildVirtualMachineImageRelationships(
...
+        // The presence of an `exactVersion` property indicates this is a gallery image.
+        //
+        // Example vm.storageProfile.imageReference:
+        // imageReference: {
+        //   "id": "/subscriptions/d3803fd6-2ba4-4286-80aa-f3d613ad59a7/resourceGroups/j1dev/providers/Microsoft.Compute/galleries/testImageGallery/images/test-image-definition",
+        //   "exactVersion": "0.0.0"
+        // }
+        //
+        // Example gallery image version "id":
+        // "/subscriptions/d3803fd6-2ba4-4286-80aa-f3d613ad59a7/resourceGroups/j1dev/providers/Microsoft.Compute/galleries/testImageGallery/images/test-image-definition/versions/0.0.0"
```
austinkelleher, line 504: In GCP, VMs can use global images as well as images that are provisioned inside of the project. Is it possible for Azure to reference images outside of the subscription?

austinkelleher (approved):

src/steps/resource-manager/compute/index.ts
ndowmon: Here's an example of what the VM's `imageReference` property looks like when using the default Ubuntu image to provision a machine: 
```
"imageReference": {
  "publisher": "Canonical",
  "offer": "UbuntuServer",
  "sku": "18.04-LTS",
  "version": "latest",
  "exactVersion": "18.04.202105120"
}
```

If the question is whether a vm in _my_ subscription could be provisioned by an image in _your_ subscription, I'm not sure. Perhaps in the future we would want to create mapped relationships in that case, but I'm not sure it's even possible.

PR #423, opened 2021-08-04
Pass subscription validation errors to client

ndowmon:
```
### Changed

- Changed `validateInvocation()` to try directly invoking
  `subscriptionClient.getSubscription(id)` when provided a `subscriptionId`
  parameter. Directly pass errors through to the user, such as:
  - `The provided subscription identifier '{{SUBSCRIPTION_ID}}' is malformed or invalid.`
  - `The subscription '{{SUBSCRIPTION_ID}}' could not be found.`
  - `The client '{{CLIENT_ID}}' with object id '{{CLIENT_ID}}' does not have authorization to perform action 'Microsoft.Resources/subscriptions/read' over scope '/subscriptions/{{SUBSCRIPTION_ID}}' or the scope is invalid.`
```


mknoedel (approved):
Great change

PR #427, opened 2021-09-02
Increase  retries from 3 to 5

ndowmon:
```
### Changed

- Increased `GraphClient` retries from 3 to 5
- Added 500ms delay between `GraphClient` retries

### CODE REVIEW

File: src/azure/graph/client.ts [hunk 1]
```diff
@@ -210,3 +217,7 @@ function getRolesFromAccessToken(accessToken: string) {
 export const testFunctions = {
   getRolesFromAccessToken,
 };
+
+function sleep(ms) {
```
austinkelleher, line 221:
```typescript
import { sleep } from "@lifeomic/attempt";
```

austinkelleher (changes requested):
Looks good! Left a small improvement comment.

src/azure/graph/client.ts
ndowmon: I did you one better.

austinkelleher (approved):

ceelias (approved):

PR #432, opened 2021-09-16
INT-1782 - Add AppServices (Section 9) CIS benchmarks

eXtremeX:
This PR adds the following (new) benchmarks:

- 9.1 Ensure App Service Authentication is set on Azure App Service (Automated)
- 9.2 Ensure web app redirects all HTTP traffic to HTTPS in Azure App Service (Automated)
- 9.3 Ensure web app is using the latest version of TLS encryption (Automated)
- 9.4 Ensure the web app has 'Client Certificates (Incoming client certificates)' set to 'On' (Automated)
- 9.5 Ensure that Register with Azure Active Directory is enabled on App Service (Automated)
- 9.6 Ensure that 'PHP version' is the latest, if used to run the web app (Manual)
- 9.7 Ensure that 'Python version' is the latest, if used to run the web app (Manual)
- 9.8 Ensure that 'Java version' is the latest, if used to run the web app (Manual)
- 9.9 Ensure that 'HTTP Version' is the latest, if used to run the web app (Manual)
- 9.10 Ensure FTP deployments are disabled (Automated)

### Some notes:
1) Types for `app.name` and `app.resourceGroup` are defined as `string | undefined` :monocle_face: . That's the reason why the entire `| undefined` part got weaved through the additional methods. My goal was, even in the unlikely event that these are indeed undefined, that nothing breaks/causes the step to error out, but for those additional properties to just be `undefined` as well.
2) The CIS benchmark didn't include an entry for `nodeVersion`, I've added it - feels useful but let me know if that property should be removed.
3) I've left some comments inside of the converter to help make sense of what's what, I can delete them before this gets merged if we don't want to have them be there.
4) For some reason Azure won't return the values of the versions (9.6-9.8). I'm confident I'm doing the correct thing with the code and furthermore even using `az` CLI tool (as shown in the CIS benchmarks examples) isn't returning those values back (we get back empty strings ""). I decided to still keep this in the code so that we still have this functionality in place if/once this gets resolved, but let me know if we want to simply ignore these benchmarks for now and delete those properties.
5) `fetchAppAuthSettings()` necessary for resolving benchmark 9.1 required required some authorization/scope modification. Without any changes, this resulted in the following error:

> Error: The client 'xxxxxxxxxxxxxxxx' with object id 'xxxxxxxxxxxxxxxx' does not have authorization to perform action 'Microsoft.Web/sites/config/list/action' over scope '/subscriptions/yyyyyyyyyy/resourceGroups/zzzzzzzzzzzzzzzz/providers/Microsoft.Web/sites/example-appservices/config/authsettings' or the scope is invalid. If access was recently granted, please refresh your credentials.

This makes me think of the following ideas:
1) We may want to note this somewhere for customers - e.g. they need to make that modification in order for this property to be successfully ingested.
2) We probably need to add some error handling for this - so that the entire step doesn't error out if this is not the case.

###  CHANGELOG.md
(to be added when this gets merged)
```
### Added

- New properties added to resources:

  | Entity                                | Properties          |
  | ------------------------------------- | ------------------- |
  | `azure_web_app`, `azure_function_app` | `authEnabled`       |
  | `azure_web_app`, `azure_function_app` | `httpsOnly`         |
  | `azure_web_app`, `azure_function_app` | `minTlsVersion`     |
  | `azure_web_app`, `azure_function_app` | `clientCertEnabled` |
  | `azure_web_app`, `azure_function_app` | `principalId`       |
  | `azure_web_app`, `azure_function_app` | `phpVersion`        |
  | `azure_web_app`, `azure_function_app` | `pythonVersion`     |
  | `azure_web_app`, `azure_function_app` | `javaVersion`       |
  | `azure_web_app`, `azure_function_app` | `nodeVersion`       |
  | `azure_web_app`, `azure_function_app` | `http20Enabled`     |
  | `azure_web_app`, `azure_function_app` | `ftpsState`         |
  
  Note: fetching app auth settings requires additional permission to be added to the custom role:
     `Microsoft.Web/sites/config/list/Action`
```

### J1QL Queries
- 9.1 Ensure App Service Authentication is set on Azure App Service (Automated)
```
# GOOD CASE
Find (azure_function_app | azure_web_app) with authEnabled = true
   
# BAD CASE
Find (azure_function_app | azure_web_app) with authEnabled = false
```
- 9.2 Ensure web app redirects all HTTP traffic to HTTPS in Azure App Service (Automated)
```
# GOOD CASE
Find (azure_function_app | azure_web_app) with httpsOnly = true
   
# BAD CASE
Find (azure_function_app | azure_web_app) with httpsOnly = false
```

- 9.3 Ensure web app is using the latest version of TLS encryption (Automated)
```
# GOOD CASE
Find (azure_function_app | azure_web_app) with minTlsVersion = "1.2"
   
# BAD CASE
Find (azure_function_app | azure_web_app) with minTlsVersion != "1.2"
```

- 9.4 Ensure the web app has 'Client Certificates (Incoming client certificates)' set to 'On' (Automated)
```
# GOOD CASE
Find (azure_function_app | azure_web_app) with clientCertEnabled = true
   
# BAD CASE
Find (azure_function_app | azure_web_app) with clientCertEnabled = false
```

- 9.5 Ensure that Register with Azure Active Directory is enabled on App Service (Automated)
```
# GOOD CASE
Find (azure_function_app | azure_web_app) with principalId != undefined
   
# BAD CASE
Find (azure_function_app | azure_web_app) with principalId = undefined
```

- 9.6 Ensure that 'PHP version' is the latest, if used to run the web app (Manual)
```
# GOOD CASE
Find (azure_function_app | azure_web_app) with phpVersion = <latest version, string>
   
# BAD CASE
Find (azure_function_app | azure_web_app) with phpVersion != <latest version, string>
```

- 9.7 Ensure that 'Python version' is the latest, if used to run the web app (Manual)
```
# GOOD CASE
Find (azure_function_app | azure_web_app) with pythonVersion = <latest version, string>
   
# BAD CASE
Find (azure_function_app | azure_web_app) with pythonVersion != <latest version, string>
```

- 9.8 Ensure that 'Java version' is the latest, if used to run the web app (Manual)
```
# GOOD CASE
Find (azure_function_app | azure_web_app) with javaVersion = <latest version, string>
   
# BAD CASE
Find (azure_function_app | azure_web_app) with javaVersion != <latest version, string>
```

- 9.9 Ensure that 'HTTP Version' is the latest, if used to run the web app (Manual)
```
# GOOD CASE
Find (azure_function_app | azure_web_app) with http20Enabled = true
   
# BAD CASE
Find (azure_function_app | azure_web_app) with http20Enabled = false
```

- 9.10 Ensure FTP deployments are disabled (Automated)
```
# GOOD CASE
Find (azure_function_app | azure_web_app) with ftpsState != "AllAllowed"
   
# BAD CASE
Find (azure_function_app | azure_web_app) with ftpsState = "AllAllowed"
```

- [Bonus, not CIS benchmark] Ensure that 'Node version' is the latest, if used to run the web app (Manual)
```
# GOOD CASE
Find (azure_function_app | azure_web_app) with nodeVersion = <latest version, string>
   
# BAD CASE
Find (azure_function_app | azure_web_app) with nodeVersion != <latest version, string>
```

Pinging @ndowmon @aiwilliams (I'm not sure if I need to tag someone else for Azure related PRs :slightly_smiling_face: )

### DIFF
--- a/src/steps/resource-manager/appservice/client.ts
+++ b/src/steps/resource-manager/appservice/client.ts
@@ -1,5 +1,10 @@
 import { WebSiteManagementClient } from '@azure/arm-appservice';
-import { AppServicePlan, Site } from '@azure/arm-appservice/esm/models';
+import {
+  AppServicePlan,
+  Site,
+  WebAppsGetAuthSettingsResponse,
+  WebAppsGetConfigurationResponse,
+} from '@azure/arm-appservice/esm/models';
 import {
   Client,
   iterateAllResources,
@@ -36,5 +41,35 @@ export class AppServiceClient extends Client {
       resourceDescription: 'appService.appServicePlans',
       callback,
     });
+  }
+
+  public async fetchAppConfiguration(
+    name: string | undefined,
+    resourceGroup: string | undefined,
+  ): Promise<WebAppsGetConfigurationResponse | null> {
+    const serviceClient = await this.getAuthenticatedServiceClient(
+      WebSiteManagementClient,
+    );
+
+    if (!name || !resourceGroup) {
+      return null;
+    }
+
+    return serviceClient.webApps.getConfiguration(resourceGroup, name);
+  }
+
+  public async fetchAppAuthSettings(
+    name: string | undefined,
+    resourceGroup: string | undefined,
+  ): Promise<WebAppsGetAuthSettingsResponse | null> {
+    const serviceClient = await this.getAuthenticatedServiceClient(
+      WebSiteManagementClient,
+    );
+
+    if (!name || !resourceGroup) {
+      return null;
+    }
+
+    return serviceClient.webApps.getAuthSettings(resourceGroup, name);
   }
 }
--- a/src/steps/resource-manager/appservice/client.ts
+++ b/src/steps/resource-manager/appservice/client.ts
@@ -43,68 +43,6 @@ export class AppServiceClient extends Client {
     });
   }
 
-  public createAppAuthConfigurationContext(): (
-    name?: string,
-    resourceGroup?: string,
-  ) => Promise<WebAppsGetAuthSettingsResponse | undefined> {
-    let canGetAuthSettings: boolean | undefined = undefined;
-
-    return async (
-      name?: string,
-      resourceGroup?: string,
-    ): Promise<WebAppsGetAuthSettingsResponse | undefined> => {
-      if (typeof canGetAuthSettings === 'undefined') {
-        canGetAuthSettings = await this.canFetchAppAuthSettings(
-          name,
-          resourceGroup,
-        );
-      }
-
-      if (!canGetAuthSettings) {
-        return;
-      }
-
-      const serviceClient = await this.getAuthenticatedServiceClient(
-        WebSiteManagementClient,
-      );
-
-      if (!name || !resourceGroup) {
-        return;
-      }
-
-      return serviceClient.webApps.getAuthSettings(resourceGroup, name);
-    };
-  }
-
-  public async canFetchAppAuthSettings(
-    name: string | undefined,
-    resourceGroup: string | undefined,
-  ): Promise<boolean | undefined> {
-    const serviceClient = await this.getAuthenticatedServiceClient(
-      WebSiteManagementClient,
-    );
-
-    if (!name || !resourceGroup) {
-      return;
-    }
-
-    try {
-      await serviceClient.webApps.getAuthSettings(resourceGroup, name);
-      return true;
-    } catch (err) {
-      this.logger.warn({ err }, 'Warning: unable to fetch app configuration.');
-      if (err.statusCode === 403 && err.code === 'AuthorizationFailed') {
-        this.logger.publishEvent({
-          name: 'MISSING_PERMISSION',
-          description:
-            'Missing permission "Microsoft.Web/sites/config/list/action", which is used to fetch WebApp Auth Settings. Please update the `JupiterOne Reader` Role in your Azure environment in order to fetch these settings for your WebApp.',
-        });
-        return false;
-      }
-      return;
-    }
-  }
-
   public async fetchAppConfiguration(
     name: string | undefined,
     resourceGroup: string | undefined,
--- a/src/steps/resource-manager/appservice/client.ts
+++ b/src/steps/resource-manager/appservice/client.ts
@@ -43,16 +43,80 @@ export class AppServiceClient extends Client {
     });
   }
 
-  public async fetchAppConfiguration(
+  public async createAppAuthConfigurationContext(): Promise<
+    (
+      name?: string,
+      resourceGroup?: string,
+    ) => Promise<WebAppsGetAuthSettingsResponse | undefined>
+  > {
+    let canGetAuthSettings: boolean | undefined = undefined;
+
+    return async (
+      name?: string,
+      resourceGroup?: string,
+    ): Promise<WebAppsGetAuthSettingsResponse | undefined> => {
+      if (typeof canGetAuthSettings === 'undefined') {
+        canGetAuthSettings = await this.canFetchAppAuthSettings(
+          name,
+          resourceGroup,
+        );
+      }
+
+      if (!canGetAuthSettings) {
+        return;
+      }
+
+      const serviceClient = await this.getAuthenticatedServiceClient(
+        WebSiteManagementClient,
+      );
+
+      if (!name || !resourceGroup) {
+        return;
+      }
+
+      return serviceClient.webApps.getAuthSettings(resourceGroup, name);
+    };
+  }
+
+  public async canFetchAppAuthSettings(
     name: string | undefined,
     resourceGroup: string | undefined,
-  ): Promise<WebAppsGetConfigurationResponse | null> {
+  ): Promise<boolean | undefined> {
     const serviceClient = await this.getAuthenticatedServiceClient(
       WebSiteManagementClient,
     );
 
     if (!name || !resourceGroup) {
-      return null;
+      return;
+    }
+
+    try {
+      await serviceClient.webApps.getAuthSettings(resourceGroup, name);
+      return true;
+    } catch (err) {
+      this.logger.warn({ err }, 'Warning: unable to fetch app configuration.');
+      if (err.statusCode === 403 && err.code === 'AuthorizationFailed') {
+        this.logger.publishEvent({
+          name: 'MISSING_PERMISSION',
+          description:
+            'Missing permission "Microsoft.Web/sites/config/list/action", which is used to fetch WebApp Auth Settings. Please update the `JupiterOne Reader` Role in your Azure environment in order to fetch these settings for your WebApp.',
+        });
+        return false;
+      }
+      return;
+    }
+  }
+
+  public async fetchAppConfiguration(
+    name: string | undefined,
+    resourceGroup: string | undefined,
+  ): Promise<WebAppsGetConfigurationResponse | undefined> {
+    const serviceClient = await this.getAuthenticatedServiceClient(
+      WebSiteManagementClient,
+    );
+
+    if (!name || !resourceGroup) {
+      return;
     }
 
     return serviceClient.webApps.getConfiguration(resourceGroup, name);
@@ -61,13 +125,13 @@ export class AppServiceClient extends Client {
   public async fetchAppAuthSettings(
     name: string | undefined,
     resourceGroup: string | undefined,
-  ): Promise<WebAppsGetAuthSettingsResponse | null> {
+  ): Promise<WebAppsGetAuthSettingsResponse | undefined> {
     const serviceClient = await this.getAuthenticatedServiceClient(
       WebSiteManagementClient,
     );
 
     if (!name || !resourceGroup) {
-      return null;
+      return;
     }
 
     return serviceClient.webApps.getAuthSettings(resourceGroup, name);
--- a/src/steps/resource-manager/appservice/client.ts
+++ b/src/steps/resource-manager/appservice/client.ts
@@ -43,12 +43,10 @@ export class AppServiceClient extends Client {
     });
   }
 
-  public async createAppAuthConfigurationContext(): Promise<
-    (
-      name?: string,
-      resourceGroup?: string,
-    ) => Promise<WebAppsGetAuthSettingsResponse | undefined>
-  > {
+  public createAppAuthConfigurationContext(): (
+    name?: string,
+    resourceGroup?: string,
+  ) => Promise<WebAppsGetAuthSettingsResponse | undefined> {
     let canGetAuthSettings: boolean | undefined = undefined;
 
     return async (
--- a/src/steps/resource-manager/appservice/converters.ts
+++ b/src/steps/resource-manager/appservice/converters.ts
@@ -5,14 +5,27 @@ import {
 } from '@jupiterone/integration-sdk-core';
 
 import { AzureWebLinker } from '../../../azure';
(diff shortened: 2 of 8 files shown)


ndowmon (changes requested):
Looks good, one additional request.

src/steps/resource-manager/appservice/client.ts
```diff
@@ -37,4 +42,34 @@ export class AppServiceClient extends Client {
...
+  ): Promise<WebAppsGetConfigurationResponse | null> {
+    const serviceClient = await this.getAuthenticatedServiceClient(
+      WebSiteManagementClient,
+    );
+
+    if (!name || !resourceGroup) {
+      return null;
+    }
+
+    return serviceClient.webApps.getConfiguration(resourceGroup, name);
```
ndowmon, line 58: In this client call, it's possible that users do not have the required permissions in order to fetch these settings. We definitely don't want to cause the entire step to fail in that case. See below for an example of what you can do to avoid this.

```suggestion
    try {
      const response = await serviceClient.webApps.getConfiguration(resourceGroup, name);
      return response;
    } catch (err) {
      this.logger.warn({ err }, 'Warning: unable to fetch app configuration.');
      if (isMissingGetAppConfigPermission(err)) {
        this.logger.publishEvent({ name: 'MISSING_PERMISSION', description: 'Missing permission <Microsoft.Whatever/whatever/whatever>, which is used to fetch WebApp Configuration. Please update the `JupiterOne Reader` Role in your Azure environment in order to fetch these settings for your WebApp Configuration.' });
      }
      return;
  }
```

In addition, the code above will have you hitting this endpoint for every web app, even if it's failing with an auth error. You probably could refactor to just check once if there is a permissions error. If there is, you can skip all of these calls in `index.ts::fetchApps()`.

src/steps/resource-manager/appservice/client.ts
```diff
@@ -37,4 +42,34 @@ export class AppServiceClient extends Client {
...
+    const serviceClient = await this.getAuthenticatedServiceClient(
+      WebSiteManagementClient,
+    );
+
+    if (!name || !resourceGroup) {
+      return null;
+    }
+
+    return serviceClient.webApps.getAuthSettings(resourceGroup, name);
+  }
```
ndowmon, line 74: In this client call, it's possible that users do not have the required permissions in order to fetch these settings. We definitely don't want to cause the entire step to fail in that case. See below for an example of what you can do to avoid this.

```suggestion
    try {
      const response = await serviceClient.webApps.getAuthSettings(resourceGroup, name);
      return response;
    } catch (err) {
      this.logger.warn({ err }, 'Warning: unable to fetch app auth settings.');
      if (isMissingGetAppAuthSettingsPermission(err)) {
        this.logger.publishEvent({ name: 'MISSING_PERMISSION', description: 'Missing permission <Microsoft.Whatever/whatever/whatever>, which is used to fetch WebApp Auth Settings. Please update the `JupiterOne Reader` Role in your Azure environment in order to fetch these settings for your WebApp Auth Settings.' });
      }
      return;
  }
```

In addition, the code above will have you hitting this endpoint for every web app, even if it's failing with an auth error. You probably could refactor to just check once if there is a permissions error. If there is, you can skip all of these calls in `index.ts::fetchApps()`.

src/steps/resource-manager/appservice/client.ts
```diff
@@ -37,4 +42,34 @@ export class AppServiceClient extends Client {
...
+  public async fetchAppAuthSettings(
+    name: string | undefined,
+    resourceGroup: string | undefined,
+  ): Promise<WebAppsGetAuthSettingsResponse | null> {
+    const serviceClient = await this.getAuthenticatedServiceClient(
+      WebSiteManagementClient,
+    );
+
+    if (!name || !resourceGroup) {
+      return null;
```
ndowmon, line 70:
```suggestion
      return;
```

src/steps/resource-manager/appservice/client.ts
```diff
@@ -37,4 +42,34 @@ export class AppServiceClient extends Client {
...
+    }
+
+    return serviceClient.webApps.getConfiguration(resourceGroup, name);
+  }
+
+  public async fetchAppAuthSettings(
+    name: string | undefined,
+    resourceGroup: string | undefined,
+  ): Promise<WebAppsGetAuthSettingsResponse | null> {
```
ndowmon, line 64:
```suggestion
  ): Promise<WebAppsGetAuthSettingsResponse | undefined> {
```

src/steps/resource-manager/appservice/client.ts
```diff
@@ -37,4 +42,34 @@ export class AppServiceClient extends Client {
...
+  public async fetchAppConfiguration(
+    name: string | undefined,
+    resourceGroup: string | undefined,
+  ): Promise<WebAppsGetConfigurationResponse | null> {
+    const serviceClient = await this.getAuthenticatedServiceClient(
+      WebSiteManagementClient,
+    );
+
+    if (!name || !resourceGroup) {
+      return null;
```
ndowmon, line 55:
```suggestion
      return;
```

src/steps/resource-manager/appservice/client.ts
```diff
@@ -37,4 +42,34 @@ export class AppServiceClient extends Client {
       callback,
     });
   }
+
+  public async fetchAppConfiguration(
+    name: string | undefined,
+    resourceGroup: string | undefined,
+  ): Promise<WebAppsGetConfigurationResponse | null> {
```
ndowmon, line 49:
```suggestion
  ): Promise<WebAppsGetConfigurationResponse | undefined> {
```

src/steps/resource-manager/appservice/converters.ts
```diff
@@ -5,14 +5,27 @@ import {
...
+  data,
+  metadata,
+  appConfig,
+  appAuthSettings,
+}: {
+  webLinker: AzureWebLinker;
+  data: Site;
+  metadata: StepEntityMetadata;
+  appConfig: WebAppsGetConfigurationResponse | null;
+  appAuthSettings: WebAppsGetAuthSettingsResponse | null;
```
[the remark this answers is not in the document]
ndowmon:
```suggestion
  appConfig: WebAppsGetConfigurationResponse | undefined;
  appAuthSettings: WebAppsGetAuthSettingsResponse | undefined;
```

ndowmon:
> 5. `fetchAppAuthSettings()` necessary for resolving benchmark 9.1 required required some authorization/scope modification. Without any changes, this resulted in the following error:
> 
> > Error: The client 'xxxxxxxxxxxxxxxx' with object id 'xxxxxxxxxxxxxxxx' does not have authorization to perform action 'Microsoft.Web/sites/config/list/action' over scope '/subscriptions/yyyyyyyyyy/resourceGroups/zzzzzzzzzzzzzzzz/providers/Microsoft.Web/sites/example-appservices/config/authsettings' or the scope is invalid. If access was recently granted, please refresh your credentials.
> 
> This makes me think of the following ideas:
> 
> 1. We may want to note this somewhere for customers - e.g. they need to make that modification in order for this property to be successfully ingested.
> 2. We probably need to add some error handling for this - so that the entire step doesn't error out if this is not the case.

Great callout @eXtremeX. This should be brought to the attention of users in two places:
1. Update `docs/jupiterone.md` documentation. The docs describe creating a custom role called `JupiterOne Reader` with a specific permission applied. You should add the additional permission there.
2. Make a note in `CHANGELOG.md`. That will remind us on version / deploy to get this messaging out to users.

As for error handling, you're 100% right. I've just added some code suggestions to the PR.

src/steps/resource-manager/appservice/client.ts
```diff
@@ -37,4 +42,96 @@ export class AppServiceClient extends Client {
...
+    }
+
+    try {
+      await serviceClient.webApps.getAuthSettings(resourceGroup, name);
+      return true;
+    } catch (err) {
+      this.logger.warn({ err }, 'Warning: unable to fetch app configuration.');
+      if (err.statusCode === 403 && err.code === 'AuthorizationFailed') {
+        this.logger.publishEvent({
+          name: 'MISSING_PERMISSION',
```
ndowmon, line 98: We use the convention of lowercase events.
```suggestion
          name: 'missing_permission',
```

src/steps/resource-manager/appservice/client.ts
ndowmon: Does this endpoint not require `Microsoft.Web/sites/config/list/action`? I thought it did but I could be wrong.

src/steps/resource-manager/appservice/client.ts
```diff
@@ -37,4 +42,96 @@ export class AppServiceClient extends Client {
...
+    );
+
+    if (!name || !resourceGroup) {
+      return;
+    }
+
+    return serviceClient.webApps.getConfiguration(resourceGroup, name);
+  }
+
+  public async fetchAppAuthSettings(
```
ndowmon, line 123: You seem to have updated the code such that this method is no longer used? This may still be a better strategy than `createAppAuthConfigurationContext`, but one or the other is going to need to get removed.

src/steps/resource-manager/appservice/client.ts
```diff
@@ -37,4 +42,96 @@ export class AppServiceClient extends Client {
...
+
+    return async (
+      name?: string,
+      resourceGroup?: string,
+    ): Promise<WebAppsGetAuthSettingsResponse | undefined> => {
+      if (typeof canGetAuthSettings === 'undefined') {
+        canGetAuthSettings = await this.canFetchAppAuthSettings(
+          name,
+          resourceGroup,
+        );
```
ndowmon, line 60: I don't think this is going to work; what if the user has for some reason granted this permission to the first WebApp that J1 encounters (setting `canGetAuthSettings: true`), but all of the other WebApps have _not_ granted this permission? This method will fail many times without alerting the user that they are missing some settings.

ndowmon: I think it would make sense to go back to the simple `client.fetchAppAuthSettings` method and do some logic in the step itself to check for this.

src/steps/resource-manager/appservice/index.ts
```diff
@@ -30,10 +30,27 @@ export async function fetchApps(
...
   await client.iterateApps(async (app) => {
+    const appAuthSettings = await fetchAppAuthConfiguration(
+      app.name,
+      app.resourceGroup,
+    );
+    const appConfig = await client.fetchAppConfiguration(
+      app.name,
+      app.resourceGroup,
+    );
+
```
ndowmon, line 44: I think it's best to keep hitting these endpoints until you encounter a `403` on either one.

```suggestion  
  let canFetchAppAuthConfiguration = true;

  await client.iterateApps(async (app) => {
    let appAuthSettings: WebAppsGetAuthSettingsResponse | undefined;
    let appConfig: WebAppsGetConfigurationResponse | undefined;
    if (canFetchAppAuthConfiguration) {
      try {
        appAuthSettings = await client.fetchAppAuthSettings(
          app.name,
          app.resourceGroup,
        );
        appConfig = await client.fetchAppConfiguration(
          app.name,
          app.resourceGroup,
        );
      } catch (err) {
        logger.warn({ err, name: app.name, resourceGroup: app.resourceGroup }, 'Warning: unable to fetch app configuration.');
        if (err.statusCode === 403 && err.code === 'AuthorizationFailed') {
          logger.publishEvent({
            name: 'missing_permission',
            description:
              'Missing permission "Microsoft.Web/sites/config/list/action", which is used to fetch WebApp Auth Settings. Please update the `JupiterOne Reader` Role in your Azure environment in order to fetch these settings for your WebApp.',
          });
          canFetchAppAuthConfiguration = false;
        }
      }
    }

```

PR #438, opened 2021-09-30
INT-1766 - Add 3.11 CIS benchmarks

eXtremeX:
**Note: this PR is rebased against the #436 PR - best if this is reviewed after #436.**

This PR adds the following (new) benchmarks:

- 3.11 Ensure Storage logging is enabled for Table service for read, write, and delete requests (Manual)

### Some notes:
1) We had to make a custom HTTP request in order to get TableProperties as seen [here](https://github.com/JupiterOne/graph-azure/compare/main...Creativice-Oy:feature/cis-3.11#diff-48f3520a72f14591e3bb6655cc135cbba293b21a11fed71a02c7ccf6c114e6f0R115). Here are some more details on our thought process that went into this:

We wanted to use [this endpoint API](https://docs.microsoft.com/en-us/rest/api/storageservices/get-table-service-properties#see-also) which is implemented through the JS client (@azure/arm-storage) as [this method](https://docs.microsoft.com/en-us/javascript/api/@azure/arm-storage/tableservices?view=azure-node-latest#getServiceProperties_string__string__msRest_RequestOptionsBase_).
Now for some reason, the method only returns [this](https://docs.microsoft.com/en-us/javascript/api/@azure/arm-storage/tableserviceproperties?view=azure-node-latest) as a response.
It only contains CORS rules and there is no information about the logging configuration whereas, on the REST API, there is.
The interesting thing is in the method description, it explicitly states that it:

> Gets the properties of a storage account’s Table service, including properties for Storage Analytics and CORS (Cross-Origin Resource Sharing) rules.

However, we were only able to see CORS rules by using it. 

###  CHANGELOG.md
(to be added when this gets merged)
```
### Added

- New properties added to resources:

  | Entity                   | Properties                             |
  | ------------------------ | -------------------------------------- |
  | `azure_storage_account`  | `tableAnalyticsLoggingReadEnabled`     |
  | `azure_storage_account`  | `tableAnalyticsLoggingWriteEnabled`    |
  | `azure_storage_account`  | `tableAnalyticsLoggingDeleteEnabled`   |

```

### J1QL Queries
- 3.11 Ensure Storage logging is enabled for Table service for read, write, and delete requests (Manual)
```
# GOOD CASE
find azure_storage_account with tableAnalyticsLoggingReadEnabled = true and tableAnalyticsLoggingWriteEnabled = true and tableAnalyticsLoggingDeleteEnabled = true
   
# BAD CASE
find azure_storage_account with tableAnalyticsLoggingReadEnabled != true or tableAnalyticsLoggingWriteEnabled != true or tableAnalyticsLoggingDeleteEnabled != true
```

Pinging @ndowmon @aiwilliams


src/steps/resource-manager/storage/client.ts
```diff
@@ -84,6 +111,45 @@ export function createStorageAccountServiceClient(options: {
...
+              headers: {
+                Authorization: `Bearer ${token?.token}`,
+                'x-ms-date': formatRFC7231(new Date()),
+                'x-ms-version': '2020-10-02',
+              },
+            },
+          );
+
+          const res = await response.text();
+          return parseStringPromise(res);
```
ndowmon, line 141: Was it possible to use the patterns used above? I.e. 

```
        const client = new TableServiceClient(
          `https://${storageAccount.name}.queue.core.windows.net`,
          credential,
        );
        try {
          const response = await client.getProperties();
          return response._response.parsedBody;
```

I see that there is a `TableServiceClient` exported from `@azure/data-tables`, but I haven't tried it out.

ndowmon (commented):
Just one question about using the `@azure/data-tables` package here. You may have already investigated that package though

src/steps/resource-manager/storage/client.ts
RPGPH: I was actually looking at `@azure/arm-storage`'s [`TableServices.getServiceProperties`](https://docs.microsoft.com/en-us/javascript/api/@azure/arm-storage/tableservices?view=azure-node-latest#getServiceProperties_string__string__msRest_RequestOptionsBase_) for this endpoint. The strange thing is it [only returns `cors` properties](https://docs.microsoft.com/en-us/javascript/api/@azure/arm-storage/tableserviceproperties?view=azure-node-latest) as the response so we opted to interface with the HTTP REST endpoint instead. I'll look into `@azure/data-tables` if we can use it for Storage Account's table services for this benchmark.

ndowmon: Thanks for checking! it's great that you got this to work with a direct `fetch()`, and that should work just fine, but leveraging Microsoft's API clients / types is always the safest way to go when it's an option.

PR #445, opened 2021-10-04
open-source managed questions

ndowmon:
We'll need to get `JUPITERONE_ACCOUNT_ID` and `JUPITERONE_API_KEY` secrets added to this repo.

```
## [5.32.0] - 2021-10-04

### Added

- Open-sourced Azure managed questions
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,6 +24,9 @@ jobs:
         run: yarn
 
       - name: Run tests
+        env:
+          JUPITERONE_ACCOUNT_ID: ${{ secrets.JUPITERONE_ACCOUNT_ID }}
+          JUPITERONE_API_KEY: ${{ secrets.JUPITERONE_API_KEY }}
         run: yarn test:ci
 
       - name: Run build
@@ -71,4 +74,4 @@ jobs:
         run: |
           echo "//registry.npmjs.org/:_authToken=${NPM_AUTH_TOKEN}" > .npmrc
           yarn
-          npm publish
+          npm publish ./dist
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,8 +25,8 @@ jobs:
 
       - name: Run tests
         env:
-          JUPITERONE_ACCOUNT_ID: ${{ secrets.JUPITERONE_ACCOUNT_ID }}
-          JUPITERONE_API_KEY: ${{ secrets.JUPITERONE_API_KEY }}
+          MANAGED_QUESTIONS_JUPITERONE_ACCOUNT_ID: ${{ secrets.MANAGED_QUESTIONS_JUPITERONE_ACCOUNT_ID }}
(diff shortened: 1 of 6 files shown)


.github/workflows/build.yml
```diff
@@ -71,4 +74,4 @@ jobs:
         run: |
           echo "//registry.npmjs.org/:_authToken=${NPM_AUTH_TOKEN}" > .npmrc
           yarn
-          npm publish
+          npm publish ./dist
```
mknoedel, line 77: Are these changes going to need to be done on every project we want to open source questions for?

jupiterone/questions/questions.yaml
```diff
@@ -0,0 +1,2149 @@
+---
```
mknoedel, line 1: There is a lot of work involved in these questions. Are we sure we want to open source them?

package.json
```diff
@@ -2,30 +2,35 @@
   "name": "@jupiterone/graph-azure",
   "version": "5.31.7",
   "description": "A graph conversion tool for https://azure.microsoft.com/.",
-  "main": "dist/index.js",
+  "main": "src/index.js",
+  "types": "src/index.d.ts",
```
mknoedel, line 6: I believe this breaks `yarn link` for deploying to dev.

.github/workflows/build.yml
ndowmon: Actually, this follows a convention that we've been using for the last few months (see [this pr](https://github.com/JupiterOne/integration-template/pull/41) in `integration-template`). But I noticed that the `graph-google-cloud` project had been updated to use this pattern, which I think is better overall. I don't think it's necessarily required though, it could be done another way.

jupiterone/questions/questions.yaml
ndowmon: The main reason for this change is to have devs start to contribute questions as code (and allow reviewers to comment directly on questions). For example, Stefan and team currently have 7 open PRs with 23 managed questions, which I would like to have in code so that we can review & discuss those lines as we would the rest of the code base.

Are your concerns intellectual property related? As in, we have worked hard on these and don't necessarily want to just give them out?

package.json
ndowmon: True, this will require users to `cd` into `dist` before running `yarn link`.

.github/workflows/build.yml
aiwilliams: I've been moving to this new pattern as well when working on integrations.

aiwilliams (approved):

aiwilliams:
> We'll need to get `JUPITERONE_ACCOUNT_ID` and `JUPITERONE_API_KEY` secrets added to this repo.

Looks like these organization secrets were added. 👍

jupiterone/questions/questions.yaml
mknoedel: Yeah, my thought is that we could leave these questions in the managed project and allow other users to add questions in the graph project.

mknoedel (approved):

PR #471, opened 2021-12-23
INT-2284 - Add status and normalize active fields onto azure_vm entity

eXtremeX:
This uses [Instance View](https://docs.microsoft.com/en-us/rest/api/compute/virtual-machines/instance-view) to get the necessary data.

The `status` field will either have `running`, `stopped` or `undefined` as its value depending on the state and if the fetching of the instance view was successful.

`active` field will be `true` or `false` depending if the `status` is `running` or `stopped`, but I've also made it so `active` is `undefined` in case `status` is `undefined` as well, that way we wouldn't be potentially falsely setting it to true/false when we "don't really know the state".

I was always been able to get the instance view and the status value while working on this task so the undefined behavior mentioned above is mostly just a safety net.


src/steps/resource-manager/compute/converters.ts
```diff
@@ -70,6 +91,8 @@ export function createVirtualMachineEntity(
     region: data.location,
     resourceGroup: resourceGroupName(data.id),
     state: data.provisioningState,
+    status,
+    active: status ? status === "running" : undefined,
```
ceelias, line 95: If the status is "stopped" do we want `active: true`?

ceelias (approved):
Looks solid!

austinkelleher (commented):
@eXtremeX This looks good, but I think we should rename the property to `state` to match `aws_instance`. I've made a change to the data-model, but updating your property doesn't necessarily need to wait for that.

See: https://github.com/JupiterOne/data-model/pull/144/files

src/steps/resource-manager/compute/index.ts
```diff
@@ -215,7 +217,17 @@ export async function fetchVirtualMachines(
   const client = new ComputeClient(instance.config, logger);
 
   await client.iterateVirtualMachines(async (vm) => {
-    const virtualMachineEntity = createVirtualMachineEntity(webLinker, vm);
+    let instanceView: VirtualMachinesInstanceViewResponse | undefined = undefined;
+    try {
+      instanceView = await client.fetchInstanceView(vm.name, getResourceGroupName(vm.id || ""));
+    } catch (err) {
+      logger.warn(
```
austinkelleher, line 224: Is this a permission boundary case?

austinkelleher (approved):

src/steps/resource-manager/compute/index.ts
eXtremeX: It's been some time since I worked on this, but from what I remember I didn't run into any issues with fetching instanceView(s) nor did I require any specific/custom permissions to be set.

That was set up so if any of the running instances ever have any troubles with fetch instanceView, we will be able to see and look into it further. It also shouldn't make entire step function break down since instanceView *just* give us additional property but doesn't otherwise break anything.

You think that's fine for the reasons above or would you change it?

PR #476, opened 2022-01-27
INT-2169 - Suppress ContainerOperationFailure Account is disabled issue

eXtremeX:


src/steps/resource-manager/storage/client.ts
```diff
@@ -309,9 +309,12 @@ export class StorageClient extends Client {
       await cb();
     } catch (e) {
       if (
+        // TODO is it possible to skip these calls altogether? How can we anticipate `FeatureNotSupportedForAccount` or `AccountIsDisabled`?
         ['FeatureNotSupportedForAccount', 'AccountIsDisabled'].includes(
           e.status,
-        ) // TODO is it possible to skip these calls altogether? How can we anticipate `FeatureNotSupportedForAccount` or `AccountIsDisabled`?
+        ) ||
+        (e.status.includes('ContainerOperationFailure') &&
+          e.status.includes('account is disabled'))
```
ndowmon, line 317: Why not just add these to the existing check?

```ts
[
  'FeatureNotSupportedForAccount', 
  'AccountIsDisabled',
  'ContainerOperationFailure',
  'account is disabled',
].includes(e.status)
```

ndowmon (approved):
Looks good to me @eXtremeX , though I added a question.

src/steps/resource-manager/storage/client.ts
eXtremeX: Hey thanks for looking into this!

I just wanted to be extra careful, for example the error we're dealing with right now is: "ContainerOperationFailure: the specified account is disabled".

If I were to just add those additional checks in that array, it could pick up (wrongly) some other errors such as: "ContainerOperationFailure: [something else is very wrong]", since it wouldn't force the `ContainerOperationFailure` and `account is disabled` to both be part of the same error.

For what it's worth I'm not sure what are all the errors that `ContainerOperationFailure` can include, this was just one slightly-lengthy (code-wise) way of making sure to be as precise as I could at this point.

Let me know if that made sense and if you think I should change it.

ndowmon: Oh, I see. Why not just check for the full string `ContainerOperationFailure: the specified account is disabled` ?

ndowmon: BTW I would love if there were some way we could determine whether the account is disabled before even trying to hit this endpoint.

eXtremeX: Pretty much similar reasons - I didn't want to risk having some additional space, character, etc result in a false negative. As I wasn't able to get the same error to appear for me locally (mentioned this on Slack) I only have the full error message from Sentry event details and this just feels safer.

eXtremeX: I could change it if you'd want me to, but I'd be worried of it not matching exactly and similar.

ndowmon: OK this is fine then.

I definitely do want to go back at some point and be able to prevent us from even making these three API calls before getting error responses.

ndowmon (approved):

PR #484, opened 2022-03-01
Fixing Cosmosdb account bug

jfish7:
added a 'function' property to the cosmosdb step to conform with the 'Service' class

example error:
```
(stepId=rm-cosmosdb-sql-databases, errorId=01aa7a34-9efb-41d5-81ea-6d72377fa61e)
  Error: Entity fails to validate as class 'Service':

  [
    {
      "instancePath": "",
      "schemaPath": "#/allOf/1/required",
      "keyword": "required",
      "params": {
        "missingProperty": "function"
      },
      "message": "must have required property 'function'"
    }
  ]
      at Object.validateEntityWithSchema (/Users/jfisher/Tools/graph-azure/node_modules/@jupiterone/data-model/src/validateEntityWithSchema.ts:15:13)
      at Object.createIntegrationEntity (/Users/jfisher/Tools/graph-azure/node_modules/@jupiterone/integration-sdk-core/src/data/createIntegrationEntity.ts:103:5)
      at Object.createAccountEntity (/Users/jfisher/Tools/graph-azure/src/steps/resource-manager/cosmosdb/converters.ts:23:10)
      at /Users/jfisher/Tools/graph-azure/src/steps/resource-manager/cosmosdb/index.ts:38:29
      at Object.iterateAllResources (/Users/jfisher/Tools/graph-azure/src/azure/resource-manager/client.ts:366:15)
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (internal/process/task_queues.js:95:5)
      at Object.fetchCosmosDBSqlDatabases [as executionHandler] (/Users/jfisher/Tools/graph-azure/src/steps/resource-manager/cosmosdb/index.ts:37:3)
```

### DIFF
--- a/src/steps/resource-manager/cosmosdb/converters.ts
+++ b/src/steps/resource-manager/cosmosdb/converters.ts
@@ -27,6 +27,7 @@ export function createAccountEntity(
         _key: data.id!,
         _type: RM_COSMOSDB_ACCOUNT_ENTITY_TYPE,
         _class: RM_COSMOSDB_ACCOUNT_ENTITY_CLASS,
+        function: ['cosmosdb-account'],
         webLink: webLinker.portalResourceUrl(data.id),
         region: normalizeLocation(data.location),
         resourceGroup: resourceGroupName(data.id),
--- a/src/steps/resource-manager/cosmosdb/converters.ts
+++ b/src/steps/resource-manager/cosmosdb/converters.ts
@@ -27,7 +27,7 @@ export function createAccountEntity(
         _key: data.id!,
         _type: RM_COSMOSDB_ACCOUNT_ENTITY_TYPE,
         _class: RM_COSMOSDB_ACCOUNT_ENTITY_CLASS,
-        function: ['cosmosdb-account'],
+        function: ['database'],
         webLink: webLinker.portalResourceUrl(data.id),
         region: normalizeLocation(data.location),
         resourceGroup: resourceGroupName(data.id),
--- a/src/steps/resource-manager/utils/createResourceGroupResourceRelationship.test.ts
+++ b/src/steps/resource-manager/utils/createResourceGroupResourceRelationship.test.ts
@@ -92,7 +92,7 @@ describe('#createResourceGroupResourceRelationship', () => {
       entities: [],
     });
 
-    const loggerErrorSpy = spyOn(context.logger, 'error');
+    const loggerErrorSpy = jest.spyOn(context.logger, 'error');
 
     const resourceEntity: Entity = {
       _class: ['Service'],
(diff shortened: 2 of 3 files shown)


src/steps/resource-manager/cosmosdb/converters.ts
```diff
@@ -27,6 +27,7 @@ export function createAccountEntity(
         _key: data.id!,
         _type: RM_COSMOSDB_ACCOUNT_ENTITY_TYPE,
         _class: RM_COSMOSDB_ACCOUNT_ENTITY_CLASS,
+        function: ['cosmosdb-account'],
```
austinkelleher, line 30: This function is required as part of the `Service` class in our `data-model`. `function` is supposed to be serve as a generic set of values, so that you can ask a question like:

> Which of my service entities serve the "auditing" function?

The exact description for `function` directly from the data-model:

> The functions provided by the service, e.g. access-review, database, load-balancing, other

See here for examples: https://github.com/JupiterOne/data-model/blob/main/src/schemas/Service.json#L32

According to the [Azure documentation](https://docs.microsoft.com/en-us/azure/cosmos-db/account-databases-containers-items):

> The Azure Cosmos account is the fundamental unit of global distribution and high availability. 

What are you thoughts on assigning a value like `['database']`?

austinkelleher (changes requested):
Hey @jfish7, thanks so much the contribution! 🥳 

I suggested an improvement to the `function` value. Additionally, a test is failing due to the new added property. You'll need to update the snapshot using `yarn jest src/steps/resource-manager/cosmosdb/converters.test.ts -u`.

FYI: You can actually run this integration without schema validation, but leveraging the 
`--disable-schema-validation` flag. See: https://github.com/JupiterOne/sdk/blob/main/docs/integrations/development.md#option---disable-schema-validation-or--v

Example:

```
yarn start --disable-schema-validation
```

Or:

```
yarn j1-integration collect --disable-schema-validation
```

src/steps/resource-manager/cosmosdb/converters.ts
jfish7: @austinkelleher thanks for the quick review. I was originally unclear as to what the `function` meant in an entity, and I chose `cosmosdb-account` to match `service-account` (which is not the best comparison). I will switch the value to `[database]` as suggested.

src/steps/resource-manager/utils/createResourceGroupResourceRelationship.test.ts
```diff
@@ -92,7 +92,7 @@ describe('#createResourceGroupResourceRelationship', () => {
       entities: [],
     });
 
-    const loggerErrorSpy = spyOn(context.logger, 'error');
+    const loggerErrorSpy = jest.spyOn(context.logger, 'error');
```
austinkelleher, line 95: 👍

austinkelleher (approved):
Looks great! Thanks so much @jfish7!

PR #495, opened 2022-03-18
INT-2806 Create mapped relationships for subscription/location

eXtremeX:


src/steps/resource-manager/network/index.test.ts
```diff
@@ -1714,23 +1714,26 @@ describe('rm-network-location-watcher-relationships', () => {
...
-      ).toMatchDirectRelationshipSchema({
-        schema: {
-          properties: {
-            _type: {
-              const: NetworkRelationships.LOCATION_HAS_NETWORK_WATCHER._type,
-            },
-          },
-        },
-      });
+        mappedRelationships.every(
```
mdaum, line 1731: note that you can use `expect(mappedRelationships).toTargetEntities` to verify your mapped relationships. [Example](https://github.com/JupiterOne/graph-veracode-v2/blob/main/src/steps/findings/index.test.ts#L63)

src/steps/resource-manager/subscriptions/index.test.ts
```diff
@@ -136,32 +132,22 @@ describe('rm-subscription-locations', () => {
...
-    expect(subscriptionLocationRelationships).toMatchDirectRelationshipSchema({
-      schema: {
-        properties: {
-          _type: { const: relationships.SUBSCRIPTION_USES_LOCATION._type },
-        },
-      },
-    });
+    expect(subscriptionLocationMappedRelationships.length).toBeGreaterThan(0);
+    expect(
+      subscriptionLocationMappedRelationships.every(
```
mdaum, line 140: note that you can use `expect(subscriptionLocationMappedRelationships).toTargetEntities` to verify your mapped relationships. [Example](https://github.com/JupiterOne/graph-veracode-v2/blob/main/src/steps/findings/index.test.ts#L63)

mdaum (approved):
small suggestion in testing, but looks good otherwise

src/steps/resource-manager/network/index.test.ts
eXtremeX: Thanks for the suggestion @mdaum, have updated the code with something similar.

neurovelho (approved):

PR #504, opened 2022-04-22
Add mgmt group -> subscription mapped relationship metadata

ndowmon:
Although we have long been creating relationships between `azure_management_group` and `azure_subscription`, the SDK recently added support for documenting mapped relationships produced by integration steps. 

---
### Mapped Relationships

The following mapped relationships are created:

| Source Entity `_type`    | Relationship `_class` | Target Entity `_type`  | Direction |
| ------------------------ | --------------------- | ---------------------- | --------- |
| `azure_management_group` | **HAS**               | `*azure_subscription*` | FORWARD   |

### DIFF
--- a/jupiterone/questions/questions.yaml
+++ b/jupiterone/questions/questions.yaml
@@ -2,6 +2,25 @@
 sourceId: managed:azure
 integrationDefinitionId: '${integration_definition_id}'
 questions:
+  - id: integration-question-azure-tenant-member-subscriptions
+    title: Which azure subscriptions are members of the same tenant?
+    description:
+      Returns a graph showing the relationships between azure_account (tenant), azure_management_group, and azure_subscription. Captures all subscriptions that are nested under three or fewer levels of management groups.
+    queries:
+      - name: Subscriptions that are part of the same tenant
+        resultsAre: Informative
+        query: |
+          FIND
+            azure_subscription 
+            (THAT HAS azure_management_group)?
+            (THAT HAS azure_management_group)?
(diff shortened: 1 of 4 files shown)


jupiterone/questions/questions.yaml
```diff
@@ -2,6 +2,25 @@
 sourceId: managed:azure
 integrationDefinitionId: '${integration_definition_id}'
 questions:
+  - id: integration-question-azure-tenant-member-subscriptions
```
ndowmon, line 5: Also added this managed question. Please note that we use the `azure_account` entity as an analog to the Tenant.

Note that Azure subscriptions are managed under "Management groups", which can be nested (similar to google cloud folders). Azure always creates a top-level default "Tenant Root Group", which relates back to the `azure_account` entity.

Screen Shot 2022-04-22 at 10 20 17 AM

JakeFerrero (dismissed):

jupiterone/questions/questions.yaml
```diff
@@ -2,6 +2,25 @@
...
+    title: Which azure subscriptions are members of the same tenant?
+    description:
+      Returns a graph showing the relationships between azure_account (tenant), azure_management_group, and azure_subscription. Captures all subscriptions that are nested under three or fewer levels of management groups.
+    queries:
+      - name: Subscriptions that are part of the same tenant
+        resultsAre: Informative
```
ndowmon, line 11:
```suggestion
        resultsAre: INFORMATIVE
```

JakeFerrero (approved):

PR #507, opened 2022-06-01
INT-3890 Do not log reports errors to users

JakeFerrero:
The /reports/credentialUserRegistrationDetails endpoint is only available to
premium azure instances. If the instance is not premium, users should not
see the error in their job logs. They currently are because we were not
awaiting the response from the endpoint.

### DIFF
--- a/src/steps/active-directory/index.test.ts
+++ b/src/steps/active-directory/index.test.ts
@@ -175,7 +175,7 @@ describe.skip('ad-user-registration-details', () => {
       },
     });
 
-    await expect(fetchUserRegistrationDetails(context)).rejects.toThrow(
+    await expect(fetchUserRegistrationDetails(context)).resolves.not.toThrow(
       IntegrationProviderAuthorizationError,
     );
   });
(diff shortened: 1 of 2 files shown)


src/steps/active-directory/index.test.ts
```diff
@@ -155,7 +155,7 @@ describe.skip('ad-user-registration-details', () => {
     });
   });
 
-  test.only('403', async () => {
+  test('403', async () => {
```
JakeFerrero, line 158: So according to line 117, the tests for this route should be skipped. However, there is a [bug](https://github.com/facebook/jest/issues/3603) that makes `test.only` ignore `describe.skip`. So this test has been running. Running and reporting a false positive, actually. It expects that an error be thrown, however we never wanted to throw the 403 in the first place. We just happened to be throwing because we were not `await`ing.

So should I skip this test (or remove it entirely), or update it so it expects to not throw?

ndowmon (dismissed):

ndowmon (approved):

PR #515, opened 2022-06-14
Int 4181 ingest azure frontdoor entities

ndowmon:
### Added

- Added `azure_frontdoor` entities

### DIFF
--- a/src/steps/resource-manager/storage/client.test.ts
+++ b/src/steps/resource-manager/storage/client.test.ts
@@ -363,7 +363,7 @@ describe('iterateTables', () => {
     developerId: 'ndowmon1',
   };
 
-  test('all', async () => {
+  test.skip('all', async () => {
     recording = setupAzureRecording({
       directory: __dirname,
       name: 'iterateTables',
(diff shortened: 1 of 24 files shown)


src/steps/resource-manager/storage/client.test.ts
```diff
@@ -363,7 +363,7 @@ describe('iterateTables', () => {
     developerId: 'ndowmon1',
   };
 
-  test('all', async () => {
+  test.skip('all', async () => {
```
austinkelleher, line 366: Is this intentional? If so, can you please create a ticket with a TODO?

zemberdotnet:
Thanks for updating the SDK dependencies and tests. I was actually trying to do some mass upgrades and changes for the new `"graph:dependencies": "j1-integration visualize-dependencies"` command and ran into errors from the tests.

If you want you can add the command to the `package.json`, or it will get added later when I push it out to all the critical ones.

src/steps/resource-manager/storage/client.test.ts
ndowmon: https://github.com/JupiterOne/graph-azure/issues/516#issue-1271289704

PR #528, opened 2022-06-28
v5.39.1

austinkelleher:
Releases #526


src/steps/resource-manager/compute/converters.ts
```diff
@@ -101,6 +101,7 @@ export function createVirtualMachineEntity(
       data.storageProfile?.osDisk,
       data.storageProfile?.dataDisks,
     ),
+    hostname: null,
```
austinkelleher, line 104: Now required by data model

PR #529, opened 2022-07-07
INT-4467 - Add `external` property to `azure_user`

austinkelleher:
Example question + J1 query:

> Which Azure users are external?

```
FIND azure_user WITH external = true
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,12 +87,12 @@
     "@azure/ms-rest-js": "^2.0.0"
   },
   "peerDependencies": {
-    "@jupiterone/integration-sdk-core": "^8.16.0"
+    "@jupiterone/integration-sdk-core": "^8.19.0"
   },
   "devDependencies": {
-    "@jupiterone/integration-sdk-core": "^8.16.0",
-    "@jupiterone/integration-sdk-dev-tools": "^8.16.0",
-    "@jupiterone/integration-sdk-testing": "^8.16.0",
+    "@jupiterone/integration-sdk-core": "^8.19.0",
+    "@jupiterone/integration-sdk-dev-tools": "^8.19.0",
+    "@jupiterone/integration-sdk-testing": "^8.19.0",
     "@types/lodash.map": "^4.6.13",
     "@types/lodash.snakecase": "^4.1.6",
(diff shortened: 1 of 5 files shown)


src/steps/active-directory/converters.ts
```diff
@@ -104,6 +104,10 @@ export function createGroupEntity(data: Group): Entity {
   });
 }
 
+function isExternalUser(username?: string) {
+  return !!(username && username.includes('#EXT#'));
```
austinkelleher, line 108: FYI I found an example in dev that includes `#EXT` in the username.

package.json
```diff
@@ -87,12 +87,12 @@
...
-    "@jupiterone/integration-sdk-core": "^8.16.0"
+    "@jupiterone/integration-sdk-core": "^8.19.0"
   },
   "devDependencies": {
-    "@jupiterone/integration-sdk-core": "^8.16.0",
-    "@jupiterone/integration-sdk-dev-tools": "^8.16.0",
-    "@jupiterone/integration-sdk-testing": "^8.16.0",
+    "@jupiterone/integration-sdk-core": "^8.19.0",
+    "@jupiterone/integration-sdk-dev-tools": "^8.19.0",
+    "@jupiterone/integration-sdk-testing": "^8.19.0",
```
austinkelleher, line 95: Small bonus upgrades

src/steps/active-directory/converters.ts
ndowmon: why not just do `return (data.userType != 'Member')`?

austinkelleher: Are you confident that these are the same thing? If so, happy to update.

ndowmon: Yep. Just invited an external user to my Azure tenant today 😉

austinkelleher: Fixed

austinkelleher:
@zemberdotnet @VDubber @RonaldEAM @Nick-NCSU - Doing some open PR spring cleaning, but I still think this is a valuable improvement. 😬 🌷

PR #540, opened 2022-09-21
INT-5201 Batch Application Failures

edreyer1:
The error in this ticket [https://jupiterone.atlassian.net/browse/INT-5201](url) occurs when you try to list the batch applications in an account that is not linked to an auto-storage account. You are unable to create an application if the account is currently unlink to an auto-storage account but the account can be unlinked after application is created.

Steps I did to setup azure to produce this error: 

1. Create Batch Account that is linked to an auto-storage account
2. Create Batch Application
3. Select "None" when selecting a storage account under the "Storage account" tab

This adds a check that autoStorage (defined in rawData of batch account) is defined before iterating through batch applications for that account

### DIFF
--- a/src/steps/resource-manager/batch/index.test.ts
+++ b/src/steps/resource-manager/batch/index.test.ts
@@ -1,18 +1,23 @@
 import {
   fetchBatchAccounts,
-  fetchBatchApplications,
+  //fetchBatchApplications,
   fetchBatchCertificates,
   fetchBatchPools,
 } from '.';
 import {
+  executeStepWithDependencies,
   MockIntegrationStepExecutionContext,
   Recording,
 } from '@jupiterone/integration-sdk-testing';
 import { IntegrationConfig } from '../../../types';
-import { setupAzureRecording } from '../../../../test/helpers/recording';
+import {
+  setupAzureRecording,
+  getMatchRequestsBy,
+} from '../../../../test/helpers/recording';
 import { createMockAzureStepExecutionContext } from '../../../../test/createMockAzureStepExecutionContext';
 import { ACCOUNT_ENTITY_TYPE } from '../../active-directory/constants';
-import { BatchEntities } from './constants';
+import { BatchEntities, STEP_RM_BATCH_APPLICATION } from './constants';
+import { getStepTestConfigForStep } from '../../../../test/integrationInstanceConfig';
 
 let recording: Recording;
 let context: MockIntegrationStepExecutionContext<IntegrationConfig>;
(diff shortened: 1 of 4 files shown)


src/steps/resource-manager/batch/index.ts
```diff
@@ -120,28 +122,31 @@ export async function fetchBatchApplications(
...
-              _class: RelationshipClass.HAS,
-              from: batchAccountEntity,
-              to: batchApplicationEntity,
-            }),
-          );
-        },
-      );
+      const rawData = getRawData<BatchAccount>(batchAccountEntity);
+
+      if (rawData?.autoStorage) {
```
ndowmon, line 127: Excellent write-up in the PR description! I had a few follow-up questions on this point:

> You are unable to create an application if the account is currently unlink to an auto-storage account but the account can be unlinked after application is created.

1. As soon as the auto-storage account is unlinked, does this API return a `409` again?
2. What happens to the batch application after the auto-storage account is unlinked? Is it deleted, or does it stick around?
3. If the batch application is _not_ deleted after the auto-storage account is unlinked, are we still ingesting it somehow? Or is it simply ignored? Is there any way to actually ingest it in this case?

src/steps/resource-manager/batch/index.ts
```diff
@@ -120,28 +122,31 @@ export async function fetchBatchApplications(
...
-          await jobState.addRelationship(
-            createDirectRelationship({
-              _class: RelationshipClass.HAS,
-              from: batchAccountEntity,
-              to: batchApplicationEntity,
-            }),
-          );
-        },
-      );
+      const rawData = getRawData<BatchAccount>(batchAccountEntity);
```
ndowmon, line 125: nit: use `batchAccountEntity` and `batchAccount` as variable names, as described in the **Fetch Child Entities** step pattern documentation https://github.com/JupiterOne/sdk/blob/main/docs/integrations/step-patterns.md#fetch-child-entities-step-pattern

```suggestion
      const batchAccount = getRawData<BatchAccount>(batchAccountEntity);
```

src/steps/resource-manager/batch/index.test.ts
```diff
@@ -195,6 +195,30 @@ describe('step - batch applications', () => {
       _type: 'azure_batch_account',
       _class: ['Service'],
       name: 'j1devbatchaccount',
+      _rawData: [
```
ndowmon, line 198: I'd like to start replacing these legacy, hard-coded tests with tests that use `executeStepWithDependencies`. In this project, there is a test at `src/steps/resource-manager/storage/index.test.ts` that uses the newer patterns - can you use that as a reference to update the test for the **Batch Application** step?

```ts
test('rm-storage-containers', async () => {
  const stepTestConfig = getStepTestConfigForStep(steps.STORAGE_CONTAINERS);

  recording = setupAzureRecording({
    name: 'rm-storage-containers',
    directory: __dirname,
    options: {
      matchRequestsBy: getMatchRequestsBy({
        config: stepTestConfig.instanceConfig,
      }),
    },
  });

  const stepResults = await executeStepWithDependencies(stepTestConfig);
  expect(stepResults).toMatchStepMetadata(stepTestConfig);
}, 10_000);
```

ndowmon (dismissed):
Looks good! I've approved, but I also made some suggestions - I think small changes like this are good opportunities to improve our testing in this project, which is in need of refinement.

src/steps/resource-manager/batch/index.ts
edreyer1: As soon as the auto-storage account is unlinked  we start getting the 409 errors again. The application doesn't get deleted, if you relink the auto-storage account you are able to start ingest information about the applications again

src/steps/resource-manager/batch/index.test.ts
```diff
@@ -1,18 +1,23 @@
 import {
   fetchBatchAccounts,
-  fetchBatchApplications,
+  //fetchBatchApplications,
```
ndowmon, line 3:
```suggestion
```

ndowmon (approved):

PR #543, opened 2022-09-27
Fix azure_kubernetes_service -> azure_resource_group relationship

ndowmon:
The step that produces `azure_resource_group-HAS->azure_kubernetes_service` relationships was not dependent on the step that produces `azure_resource_group` entities, which prevented these relationships from being properly created. 

Adding the `rm-resources-resource-groups` step as a dependency caused the relationship to be created.

Screen Shot 2022-09-27 at 10 03 24 AM


src/steps/resource-manager/container-services/index.test.ts
```diff
@@ -1,81 +1,33 @@
...
-    const instanceConfig = configFromEnv;
-
-    recording = setupAzureRecording({
-      directory: __dirname,
-      name: 'container-services-step-clusters',
-      options: {
-        matchRequestsBy: getMatchRequestsBy({ config: configFromEnv }),
-      },
-    });
+test('rm-container-services-cluster', async () => {
```
JakeFerrero, line 17: Should this new test be _in addition_ to the old `container services clusters` test rather than replacing it? or is that step no longer applicable?

ndowmon: This test replaces that one. Using `toMatchStepDependencies`, we make the same guarantees as the last test and more (without so much manual test instrumentation).

JakeFerrero: Got it, that makes sense. Thanks.

JakeFerrero (approved):

PR #549, opened 2023-01-13
Azure FetchSubscriptions Pagination

Gonzalo-Avalos-Ribas:
Added support for pagination of the API that fetches all subscriptions from a specific Active Directory.
This allows to fetch more than just a single page of subscriptions (1000 subscriptions).
It is used by the instances that have Configure Subscription instances set on instance setup.

### DIFF
--- a/src/steps/resource-manager/subscriptions/client.ts
+++ b/src/steps/resource-manager/subscriptions/client.ts
@@ -72,20 +72,31 @@ export class J1SubscriptionClient extends Client {
         passSubscriptionId: false,
       },
     );
-    const subscriptions = await request(
-      // Need API version to 2020-01-01 in order to return subscription tags
-      // serviceClient.subscriptions.list() does not work because the api version is too old
-      // sendOperationRequest was the only way I found to change the API version with this sdk
-      async () =>
-        await serviceClient.sendOperationRequest(
-          {},
-          listSubscripsionsOperationSpec,
-        ),
-      this.logger,
-      'subscription',
-      FIVE_MINUTES,
-    );
-    return subscriptions?._response?.parsedBody;
+    let allSubscriptions: any[] = [];
+    let nextLink;
+    do {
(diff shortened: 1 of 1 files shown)


src/steps/resource-manager/subscriptions/client.ts
```diff
@@ -72,20 +72,31 @@ export class J1SubscriptionClient extends Client {
...
-        await serviceClient.sendOperationRequest(
-          {},
-          listSubscripsionsOperationSpec,
-        ),
-      this.logger,
-      'subscription',
-      FIVE_MINUTES,
-    );
-    return subscriptions?._response?.parsedBody;
+    let allSubscriptions: any[] = [];
```
gastonyelmini, line 75: This could be something like:

```
let allSubscriptions: Array<Subscription> = [];
```

gastonyelmini (commented):
Looks good! Nice job. Just a quick question.

src/steps/resource-manager/subscriptions/client.ts
Gonzalo-Avalos-Ribas: Thanks! Fixed this on next commit

gastonyelmini (approved):

src/steps/resource-manager/subscriptions/client.ts
```diff
@@ -72,20 +72,31 @@ export class J1SubscriptionClient extends Client {
...
+              : listSubscripsionsOperationSpec,
+          ),
+        this.logger,
+        'subscription',
+        FIVE_MINUTES,
+      );
+      allSubscriptions = [
+        ...allSubscriptions,
+        ...subscriptions?._response?.parsedBody,
+      ];
```
ndowmon, line 96:
```suggestion
      allSubscriptions = allSubscriptions.concat(subscriptions?._response?.parsedBody);
```

ndowmon (approved):

PR #552, opened 2023-01-23
Azure permission documentation part 1: Added permissions to each step

Gonzalo-Avalos-Ribas:
Context and motivation:
We must provide the required list of permissions that require a review of all the SDK calls that we are doing and collect all of them. As part of making this process more scalable, this spike ticket was created, in order to investigate a way to automate the process of collecting this permissions and adding this to jupiterone.md README automatically. This script might be migrated to the SDK in the future to apply this process to all services.

Permissions in azure could be referring to API permissions, or Role based permissions (related to RBAC azure support). We want to document each of them separately.

How it works
`yarn document:permissions` command will be added. This script imports invocationConfig variable from steps index to be able to collect permissions. This permissions will be added to the IntegrationStep array in each step as permissions. This would look like this:

```
{
    id: ...,
    name: ...,
    dependsOn: [],
    entities: [],
    relationships: [],
    executionHandler: () => {},
    apiPermissions: ['permission3', 'permission4'],
    rolePermissions: ['permission1', 'permission2'],
},
```

### DIFF
--- a/docs/jupiterone.md
+++ b/docs/jupiterone.md
@@ -532,3 +532,104 @@ END OF GENERATED DOCUMENTATION AFTER BELOW MARKER
 [2]:
   https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-manager-api-authentication
 [3]: https://docs.microsoft.com/en-us/graph/api/organization-get
+
+<!-- {J1_PERMISSIONS_DOCUMENTATION_MARKER_START} -->
+<!-- {J1_PERMISSIONS_DOCUMENTATION_ROLE_PERMISSIONS_START} -->
+
+| Role Permissions List                                                                           |
+| ----------------------------------------------------------------------------------------------- |
+| `Microsoft.Advisor/recommendations/read`                                                        |
+| `Microsoft.Advisor/recommendations/suppressions/read`                                           |
+| `Microsoft.ApiManagement/service/apis/read`                                                     |
+| `Microsoft.ApiManagement/service/read`                                                          |
+| `Microsoft.Authorization/classicAdministrators/read`                                            |
+| `Microsoft.Authorization/locks/read`                                                            |
+| `Microsoft.Authorization/policyAssignments/read`                                                |
+| `Microsoft.Authorization/policyDefinitions/read`                                                |
+| `Microsoft.Authorization/roleAssignments/read`                                                  |
+| `Microsoft.Authorization/roleDefinitions/read`                                                  |
+| `Microsoft.Batch/batchAccounts/applications/read`                                               |
+| `Microsoft.Batch/batchAccounts/certificates/read`                                               |
+| `Microsoft.Batch/batchAccounts/pools/read`                                                      |
+| `Microsoft.Batch/batchAccounts/read`                                                            |
+| `Microsoft.Cache/redis/firewallRules/read`                                                      |
+| `Microsoft.Cache/redis/linkedServers/read`                                                      |
+| `Microsoft.Cache/redis/read`                                                                    |
+| `Microsoft.Cdn/profiles/endpoints/read`                                                         |
+| `Microsoft.Cdn/profiles/read`                                                                   |
+| `Microsoft.Compute/disks/read`                                                                  |
+| `Microsoft.Compute/galleries/images/read`                                                       |
+| `Microsoft.Compute/galleries/images/versions/read`                                              |
+| `Microsoft.Compute/galleries/read`                                                              |
+| `Microsoft.Compute/images/read`                                                                 |
+| `Microsoft.Compute/virtualMachines/extensions/read`                                             |
+| `Microsoft.Compute/virtualMachines/read`                                                        |
+| `Microsoft.ContainerInstance/containerGroups/read`                                              |
+| `Microsoft.ContainerRegistry/registries/read`                                                   |
+| `Microsoft.ContainerRegistry/registries/webhooks/read`                                          |
+| `Microsoft.ContainerService/managedClusters/read`                                               |
+| `Microsoft.DBforMariaDB/servers/databases/read`                                                 |
+| `Microsoft.DocumentDB/databaseAccounts/sqlDatabases/read`                                       |
+| `Microsoft.EventGrid/domains/read`                                                              |
+| `Microsoft.EventGrid/domains/topics/eventSubscriptions/read`                                    |
+| `Microsoft.EventGrid/domains/topics/read`                                                       |
+| `Microsoft.EventGrid/topics/eventSubscriptions/read`                                            |
+| `Microsoft.EventGrid/topics/read`                                                               |
+| `Microsoft.Insights/ActivityLogAlerts/Read`                                                     |
+| `Microsoft.Insights/LogProfiles/Read`                                                           |
+| `Microsoft.KeyVault/vaults/keys/read`                                                           |
+| `Microsoft.KeyVault/vaults/read`                                                                |
+| `Microsoft.KeyVault/vaults/secrets/read`                                                        |
+| `Microsoft.Management/managementGroups/read`                                                    |
+| `Microsoft.Network/azurefirewalls/read`                                                         |
+| `Microsoft.Network/dnszones/read`                                                               |
+| `Microsoft.Network/dnszones/recordsets/read`                                                    |
+| `Microsoft.Network/frontDoors/backendPools/read`                                                |
+| `Microsoft.Network/frontDoors/frontendEndpoints/read`                                           |
+| `Microsoft.Network/frontDoors/read`                                                             |
+| `Microsoft.Network/frontDoors/routingRules/read`                                                |
+| `Microsoft.Network/frontDoors/rulesEngines/read`                                                |
+| `Microsoft.Network/loadBalancers/read`                                                          |
+| `Microsoft.Network/networkInterfaces/read`                                                      |
+| `Microsoft.Network/networkSecurityGroups/read`                                                  |
+| `Microsoft.Network/networkWatchers/flowLogs/read`                                               |
+| `Microsoft.Network/networkWatchers/read`                                                        |
+| `Microsoft.Network/privateDnsZones/read`                                                        |
+| `Microsoft.Network/privateDnsZones/recordsets/read`                                             |
+| `Microsoft.Network/privateEndpoints/read`                                                       |
+| `Microsoft.Network/publicIPAddresses/read`                                                      |
+| `Microsoft.Network/virtualNetworks/read`                                                        |
+| `Microsoft.OperationalInsights/workspaces/providers/Microsoft.Insights/diagnosticSettings/Read` |
+| `Microsoft.PolicyInsights/policyStates/summarize/read`                                          |
+| `Microsoft.Resources/subscriptions/locations/read`                                              |
+| `Microsoft.Resources/subscriptions/read`                                                        |
+| `Microsoft.Resources/subscriptions/resourceGroups/read`                                         |
+| `Microsoft.Security/assessments/read`                                                           |
+| `Microsoft.Security/autoProvisioningSettings/read`                                              |
+| `Microsoft.Security/pricings/read`                                                              |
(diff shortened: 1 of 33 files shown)


src/steps/active-directory/index.ts
```diff
@@ -241,6 +239,7 @@ export const activeDirectorySteps: Step<
     ],
     dependsOn: [STEP_AD_ACCOUNT, STEP_AD_USER_REGISTRATION_DETAILS],
     executionHandler: fetchUsers,
+    apiPermissions: ['User.Read'],
```
ndowmon, line 242: I don't see a `User.Read` Application permission, but I do see a `User.Read.All` permission. (There is a `User.Read` "Delegated" permission. I think our documentation claims that it's required, but I'm not sure it is.)
Screenshot 2023-01-24 at 9 03 07 PM

However, it seems to me like this step actually uses `Directory.Read.All` to list all users. Can you check?

docs/jupiterone.md
```diff
@@ -532,3 +532,104 @@ END OF GENERATED DOCUMENTATION AFTER BELOW MARKER
 [2]:
   https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-manager-api-authentication
 [3]: https://docs.microsoft.com/en-us/graph/api/organization-get
+
+<!-- {J1_PERMISSIONS_DOCUMENTATION_MARKER_START} -->
+<!-- {J1_PERMISSIONS_DOCUMENTATION_ROLE_PERMISSIONS_START} -->
+
+| Role Permissions List                                                                           |
```
ndowmon, line 539: 🔥

src/steps/resource-manager/advisor/index.ts
```diff
@@ -86,5 +80,9 @@ export const advisorSteps: Step<
       ...getResourceManagerSteps().executeFirstSteps,
     ],
     executionHandler: fetchRecommendations,
+    rolePermissions: [
+      'Microsoft.Advisor/recommendations/read',
+      'Microsoft.Advisor/recommendations/suppressions/read',
```
ndowmon, line 85: I notice that there are actually two `/suppressions/read` permissions:

1. `Microsoft.Advisor/suppressions/read`
2. `Microsoft.Advisor/recommendations/suppressions/read`

Do you know what the difference is? How did you know which one to use?

src/steps/resource-manager/databases/index.ts
```diff
@@ -38,6 +32,7 @@ export const databaseSteps: Step<
     ],
     dependsOn: [STEP_AD_ACCOUNT, STEP_RM_RESOURCES_RESOURCE_GROUPS],
     executionHandler: fetchMariaDBDatabases,
+    rolePermissions: ['Microsoft.DBforMariaDB/servers/databases/read'],
```
ndowmon, line 35: Does this also require `Microsoft.DBforMariaDB/servers/read`? 

And also maybe `Microsoft.DBforMariaDB/servers/providers/Microsoft.Insights/diagnosticSettings/read`? There are a number of these `diagnosticSettings` permissions across different steps. I wonder if a higher-level permission exists to grant `Microsoft.Insights/diagnosticSettings/read` across all? Have you looked into this?

ndowmon: Looks like the `Microsoft.<Database>/servers/read` needs to be added for a few of these database steps FYI.

src/steps/resource-manager/databases/index.ts
```diff
@@ -54,6 +49,7 @@ export const databaseSteps: Step<
     ],
     dependsOn: [STEP_AD_ACCOUNT, STEP_RM_RESOURCES_RESOURCE_GROUPS],
     executionHandler: fetchMySQLDatabases,
+    rolePermissions: ['Microsoft.DBforMariaDB/servers/databases/read'],
```
ndowmon, line 52: I believe this should be `Microsoft.DBforMySQL/servers/databases/read` rather than MariaDB.

src/steps/resource-manager/databases/index.ts
```diff
@@ -54,6 +49,7 @@ export const databaseSteps: Step<
...
   },
   ...postgreSqlSteps,
   ...sqlSteps,
```
ndowmon, line 55: There are another 3 steps in `postgreSqlSteps` and 5 in `sqlSteps` - can you make sure to add those?

src/steps/resource-manager/event-grid/index.ts
```diff
@@ -245,6 +241,7 @@ export const eventGridSteps: Step<
     ],
     dependsOn: [STEP_AD_ACCOUNT, STEP_RM_RESOURCES_RESOURCE_GROUPS],
     executionHandler: fetchEventGridDomains,
+    rolePermissions: ['Microsoft.EventGrid/domains/read'],
```
ndowmon, line 244: Might need to add `diagnosticSettings/read` again here and elsewhere.

src/steps/resource-manager/frontdoor/index.ts
```diff
@@ -161,16 +159,15 @@ async function fetchFrontendEndpoints(
...
->[] = [
+export const frontdoorSteps: AzureIntegrationStep[] = [
   {
     id: FrontDoorStepIds.FETCH_FRONTDOORS,
     name: 'Fetch FrontDoors',
     entities: [FrontDoorEntities.FRONTDOOR],
     relationships: [FrontDoorRelationships.RESOURCE_GROUP_HAS_FRONTDOOR],
     executionHandler: fetchFrontDoors,
     dependsOn: [STEP_AD_ACCOUNT, STEP_RM_RESOURCES_RESOURCE_GROUPS],
+    rolePermissions: ['Microsoft.Network/frontDoors/read'],
```
ndowmon, line 170: Surprising, but this is actually the only step that makes an API call. If you look closely at the other four steps, they make no API call. Those shouldn't need any `rolePermissions`.

src/steps/resource-manager/key-vault/index.ts
```diff
@@ -225,6 +222,7 @@ export const keyvaultSteps: Step<
     relationships: [KeyVaultRelationships.KEY_VAULT_CONTAINS_KEY],
     dependsOn: [STEP_RM_KEYVAULT_VAULTS],
     executionHandler: fetchKeyVaultKeys,
+    rolePermissions: ['Microsoft.KeyVault/vaults/keys/read'],
```
ndowmon, line 225: Keys and Secrets are a little tricky - they can be accessed either though normal RBAC, _or_ though a "Key Vault Access Policy". See https://learn.microsoft.com/en-us/azure/key-vault/general/security-features#controlling-access-to-key-vault-data

I don't think you necessarily need to do anything different here, but take a look at that documentation and let me know what you think.

src/steps/resource-manager/policy-insights/index.ts
```diff
@@ -156,16 +154,15 @@ export async function buildPolicyStateResourceRelationships(
   );
 }
 
-export const policyInsightSteps: Step<
-  IntegrationStepExecutionContext<IntegrationConfig>
->[] = [
+export const policyInsightSteps: AzureIntegrationStep[] = [
```
ndowmon, line 157: I think there's another service, `src/steps/resource-manager/policy`, that may have been missed. Looks like we make about 7 API calls in that service.

src/steps/resource-manager/policy/index.ts
```diff
@@ -145,5 +142,6 @@ export const policySteps: Step<
     ],
     dependsOn: [STEP_AD_ACCOUNT, PolicySteps.POLICY_ASSIGNMENTS],
     executionHandler: fetchPolicyDefinitionsForAssignments,
+    rolePermissions: ['Microsoft.Authorization/policyDefinitions/read'],
```
ndowmon, line 145: I notice this makes a number of different API calls. Is `Microsoft.Authorization/policySetDefinitions/read` also necessary here?

src/steps/resource-manager/security/index.ts
```diff
@@ -225,6 +222,7 @@ export const securitySteps: Step<
     relationships: [SecurityRelationships.SUBSCRIPTION_HAS_SETTING],
     dependsOn: [STEP_AD_ACCOUNT, subscriptionSteps.SUBSCRIPTION],
     executionHandler: fetchSecurityCenterSettings,
+    rolePermissions: ['Microsoft.Security/pricings/read'],
```
ndowmon, line 225: This might have been a typo? `Microsoft.Security/settings/read`

src/steps/resource-manager/storage/index.ts
```diff
@@ -439,6 +436,10 @@ export const storageSteps: Step<
     relationships: [relationships.STORAGE_ACCOUNT_HAS_FILE_SHARE],
     dependsOn: [STEP_AD_ACCOUNT, steps.STORAGE_ACCOUNTS],
     executionHandler: fetchStorageFileShares,
+    rolePermissions: [
+      'Microsoft.Storage/storageAccounts/fileServices/shares/read',
+      'Microsoft.Storage/storageAccounts/fileServices/shares/read',
```
ndowmon, line 441: Is this a duplicate?

src/steps/resource-manager/storage/index.ts
```diff
@@ -463,5 +468,6 @@ export const storageSteps: Step<
     relationships: [relationships.STORAGE_ACCOUNT_HAS_TABLE],
     dependsOn: [STEP_AD_ACCOUNT, steps.STORAGE_ACCOUNTS],
     executionHandler: fetchStorageTables,
+    rolePermissions: ['Microsoft.Storage/storageAccounts/queueServices/read'],
```
ndowmon, line 471: Might be `'Microsoft.Storage/storageAccounts/tableServices/read`

src/steps/resource-manager/storage/index.ts
```diff
@@ -431,6 +427,7 @@ export const storageSteps: Step<
       STEP_RM_KEYVAULT_VAULTS,
     ],
     executionHandler: fetchStorageAccounts,
+    rolePermissions: ['Microsoft.Storage/storageAccounts/read'],
```
ndowmon, line 430: This step also makes three kind of different API calls...

```
    const storageBlobServiceProperties = await storageAccountServiceClient.getBlobServiceProperties();
    const storageQueueServiceProperties = await storageAccountServiceClient.getQueueServiceProperties();
    const storageTableServiceProperties = await storageAccountServiceClient.getTableServiceProperties();
```
Not sure which permissions are required for these, since they're not from the typical API client.

src/types.ts
```diff
@@ -61,3 +62,9 @@ export interface IntegrationConfig extends IntegrationInstanceConfig {
 
   configureSubscriptionInstances?: boolean;
 }
+
+export interface AzureIntegrationStep
+  extends IntegrationStep<IntegrationConfig> {
+  rolePermissions?: Array<string>;
+  apiPermissions?: Array<string>;
```
ndowmon, line 69: Can you add docstrings?

```suggestion
  /**
   * Azure RBAC permissions that are required to ingest data from this step. 
   * These permissions are assigned to an IAM Role and bound to a principal 
   * through an IAM Role Binding. They are attached to Azure Subscription or 
   * Azure Management Groups.
   * 
   * See https://learn.microsoft.com/en-us/azure/role-based-access-control/
   */
  rolePermissions?: Array<string>;
  /**
   * Azure API Permissions assigned directly to a service principal within 
   * Azure Active Directory. Specifically, these permissions allow access to 
   * the Azure Graph API. 
   * 
   * See https://learn.microsoft.com/en-us/graph/use-the-api
   */
  apiPermissions?: Array<string>;
```

ndowmon (commented):
Thanks for all your detailed analysis, Gonzalo! Looks very good. 

I caught a few edge cases that I mentioned - let's connect tomorrow about them.

You may have already done so, but if not, I would suggest configuring an integration with credentials that use _only_ these permissions - you should be able to create a custom role and assign nothing else to the principal. Then execute the integration and see if any steps fail.

Thankfully it looks like you can use JSON to create a custom role, so you won't need to click through the Azure UI to add each individual permission.

PR #553, opened 2023-01-24
Azure permission documentation part 2: Added script to compile the permissions into documentation

Gonzalo-Avalos-Ribas:


husky.config.js
```diff
@@ -2,7 +2,7 @@ const { hooks } = require('@jupiterone/integration-sdk-dev-tools/config/husky');
 
 module.exports = {
   hooks: {
-    'pre-commit': `${hooks['pre-commit']} && yarn precommit`,
+    'pre-commit': `${hooks['pre-commit']} && yarn document:permissions && yarn precommit`,
```
gastonyelmini, line 5: As this is a bit different, maybe to follow the current implementation and be consistent, we could just modify the package.json `precommit` command to be something like:

```
yarn document:diagnostic-settings && yarn document:permissions && git add docs/jupiterone.md && lint-staged
```

What do you think?

docs/jupiterone.md
```diff
@@ -532,3 +532,104 @@ END OF GENERATED DOCUMENTATION AFTER BELOW MARKER
 [2]:
   https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-manager-api-authentication
 [3]: https://docs.microsoft.com/en-us/graph/api/organization-get
+
+<!-- {J1_PERMISSIONS_DOCUMENTATION_MARKER_START} -->
+<!-- {J1_PERMISSIONS_DOCUMENTATION_ROLE_PERMISSIONS_START} -->
+
+| Role Permissions List                                                                           |
```
gastonyelmini, line 539: In Google permissions pr I added the 

```
[`Permissions List (${permissionsList.length})`]
```
 just to let the user know the amount of permissions that need to be added, so that the can double check if they added all of them or just missed some of them 

https://github.com/JupiterOne/graph-google-cloud/pull/570/files

husky.config.js
Gonzalo-Avalos-Ribas: Yep, i thought about it, or maybe even adding the `yarn document:permissions `after the `yarn document:diagnostic-settings` on the `yarn precommit`. But i wasn't sure about any of them, I'll be back on this one.

docs/jupiterone.md
Gonzalo-Avalos-Ribas: This makes sense, I'll add it, thanks!

husky.config.js
ndowmon: Either works, but I'm slightly preferring @gastonyelmini's suggestion.

ndowmon (dismissed):
I left one comment, but this looks grand. Thanks for your efforts!

gastonyelmini (approved):
🚀

PR #555, opened 2023-02-02
Fix(INT-6993): Some provider API Errors where being displayed incorrectly

Gonzalo-Avalos-Ribas:
When an API error ocurred related to an AzureRestError, it was displayed as:

`Provider API failed at policyInsights.policyStates: undefined undefined`

Now we get:

`Provider API failed at policyInsights.policyStates: AuthorizationFailed The client '[...]' with object id '[...]' does not have authorization to perform action 'Microsoft.PolicyInsights/policyStates/queryResults/read' over scope '/subscriptions/[...]' or the scope is invalid. If access was recently granted, please refresh your credentials.`

### DIFF
--- a/src/azure/resource-manager/client.ts
+++ b/src/azure/resource-manager/client.ts
@@ -277,8 +277,8 @@ export async function request<T extends ResourceResponse>(
       let status = '';
       let statusText = '';
       if (err instanceof AzureRestError) {
-        status = err.body?.code;
-        statusText = err.body?.message;
+        status = err.body?.code ?? err.body?.error?.code;
+        statusText = err.body?.message ?? err.body?.error?.message;
       } else if (err instanceof FetchError) {
         status = err.code!;
         statusText = err.message;
--- a/src/azure/resource-manager/client.ts
+++ b/src/azure/resource-manager/client.ts
@@ -277,7 +277,7 @@ export async function request<T extends ResourceResponse>(
       let status = '';
       let statusText = '';
       if (err instanceof AzureRestError) {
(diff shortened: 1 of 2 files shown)


src/azure/resource-manager/client.ts
```diff
@@ -277,8 +277,8 @@ export async function request<T extends ResourceResponse>(
       let status = '';
       let statusText = '';
       if (err instanceof AzureRestError) {
-        status = err.body?.code;
-        statusText = err.body?.message;
+        status = err.body?.code ?? err.body?.error?.code;
```
ndowmon, line 280: Can you please add unit tests that assert that the message is formatted something like
```
Provider API failed at policyInsights.policyStates: AuthorizationFailed The client '[...]' with object id '[...]' does not have authorization to perform action 'Microsoft.PolicyInsights/policyStates/queryResults/read' over scope '/subscriptions/[...]' or the scope is invalid. If access was recently granted, please refresh your credentials.
```

ndowmon: Also, do we have the HTTP Status code (e.g. `403`) available in these errors? I wonder if it'd be preferable to see `AuthorizationFailed` or `403` in the `status` field.

Gonzalo-Avalos-Ribas: Will add the unit tests.
I think since the graph-aws project uses the number (e.g. 403), we should use it as well, so I'll make sure to also change that. Thanks!

ndowmon (approved):

PR #557, opened 2023-02-03
Advisor refactor

Gonzalo-Avalos-Ribas:
All the recordings an step logic was moved to an executionHandlers folder, and the test file was renamed to the specific step it was testing. Also moved to the executionHandlers file.

We could leave the client.test.ts out of the executionHandlers folder, but becouse it makes API calls, we would have 2 recording folders (I'm assuming you can't decide where the recording file is created. Can you?)


src/steps/resource-manager/advisor/executionHandlers/recommendations.ts
```diff
@@ -0,0 +1,88 @@
...
+
+export const recommendationSteps: AzureIntegrationStep[] = [
+  {
+    id: AdvisorSteps.RECOMMENDATIONS,
+    name: 'Recommendations',
+    entities: [AdvisorEntities.RECOMMENDATION],
+    relationships: [
+      AdvisorRelationships.ASSESSMENT_IDENTIFIED_FINDING,
+      AdvisorRelationships.ANY_RESOURCE_HAS_FINDING,
+    ],
```
ndowmon, line 79: Interesting... this should really be three steps:

- `fetch-recommendations`
- `build-assessment-recommendation-relationships`
- `build-recommendation-resource-relationships`

I also think that the `build-recommendation-resource-relationships` step should use the `dependencyGraphId: 'last'` argument that's used in `graph-aws` and `graph-google-cloud`. The way the `dependsOn` works now, if for some reason _any_ previous step is disabled, this step will not run.

Gonzalo-Avalos-Ribas: Thanks for the comments! First part makes a lot of sense.

I understood what you described on the second part but i'd love a little more context on the difference. 

Screenshot 2023-02-03 at 14 44 17

Im pretty sure the DOCstring assigned to the property on the step is wrong since it describes the property as an array of identifiers and it's a string

src/steps/resource-manager/advisor/executionHandlers/recommendations.ts
```diff
@@ -0,0 +1,134 @@
...
+  {
+    id: AdvisorSteps.RECOMMENDATIONS,
+    name: 'Recommendations',
+    entities: [AdvisorEntities.RECOMMENDATION],
+    relationships: [],
+    dependsOn: [
+      STEP_AD_ACCOUNT,
+      SecuritySteps.ASSESSMENTS,
+      ...getResourceManagerSteps().executeFirstSteps,
+    ],
```
ndowmon, line 113: This should only depend on the `STEP_AD_ACCOUNT` step:

```suggestion
    dependsOn: [
      STEP_AD_ACCOUNT,
    ],
```

src/steps/resource-manager/advisor/executionHandlers/recommendations.ts
```diff
@@ -0,0 +1,134 @@
...
+    executionHandler: fetchRecommendations,
+    rolePermissions: ['Microsoft.Advisor/recommendations/read'],
+  },
+  {
+    id: AdvisorSteps.ASSESSMENT_RECOMMENDATION_RELATIONSHIP,
+    name: 'Assessment to recommendation relationship',
+    entities: [],
+    relationships: [AdvisorRelationships.ASSESSMENT_IDENTIFIED_FINDING],
+    dependsOn: [AdvisorSteps.RECOMMENDATIONS],
```
ndowmon, line 122: This should depend on both `RECOMMENDATIONS` and `ASSESSMENTS`:

```suggestion
    dependsOn: [
      AdvisorSteps.RECOMMENDATIONS,
      SecuritySteps.ASSESSMENTS,
    ],
```

src/steps/resource-manager/advisor/executionHandlers/recommendations.ts
```diff
@@ -0,0 +1,134 @@
...
+    executionHandler: buildAssesmentToRecommendationRelationship,
+  },
+  {
+    id: AdvisorSteps.RESOURCE_RECOMMENDATION_RELATIONSHIP,
+    name: 'Resource to recommendation relationship',
+    entities: [],
+    relationships: [AdvisorRelationships.ANY_RESOURCE_HAS_FINDING],
+    dependsOn: [AdvisorSteps.RECOMMENDATIONS],
+    executionHandler: buildResourceToRecommendationRelationship,
+    //dependencyGraphId: 'last'
```
ndowmon, line 132: You'll want to replace `dependsOn` with `dependencyGraphId: 'last'` here.

```suggestion
    executionHandler: buildResourceToRecommendationRelationship,
    dependencyGraphId: 'last'
```

You'll also need to update the top-level `src/index.ts` file to include `dependencyGraphOrder`:

```diff
export const invocationConfig = {
  loadExecutionConfig,
  instanceConfigFields,
  validateInvocation,
  integrationSteps,
  getStepStartStates,
+ dependencyGraphOrder: ['last'],
};
```
See https://github.com/JupiterOne/graph-aws/blob/main/src/index.ts#L21 for an example

ndowmon: Once you add `dependencyGraphId` to this step, your test will probably fail. Take a look at `graph-aws` for inspiration about how to test steps with `dependencyGraphId: 'last'`

src/steps/resource-manager/advisor/executionHandlers/recommendations.test.ts
```diff
@@ -0,0 +1,72 @@
...
+//     options: {
+//       matchRequestsBy: getMatchRequestsBy({
+//         config: stepTestConfig.instanceConfig,
+//       }),
+//     },
+//   });
+
+//   const stepResults = await executeStepWithDependencies(stepTestConfig);
+//   expect(stepResults).toMatchStepMetadata(stepTestConfig);
+// }, 10_000);
```
ndowmon, line 72: small nit: prefer to use `test.skip()` rather than bulk commenting out tests. That way, test runners know that a test exists here but was skipped.

src/steps/resource-manager/advisor/executionHandlers/recommendations.test.ts
```diff
@@ -0,0 +1,72 @@
...
```
ndowmon, line 53: This test doesn't need `dependencyGraphOrder: last`... is this also failing?

ndowmon (approved):

PR #558, opened 2023-02-20
Subscription Refactor

Gonzalo-Avalos-Ribas:
This commit has a couple of things:

- Refactor of the subscriptions steps. This includes the modification some files that referenced the fetchsubscription execution handler. Also generated the recordings. It seems like the location step didn't create an entity location, so i removed it from the stepDefinition.
- Moved some tests on the advisor service that were in the incorrect folder.
- Added a mutation to the recordings file. The idea is to save the recordings files without the literal subscription Id when running `yarn test:env`, but rather save the string 'subscription-id' instead. This is easily removable if necessary, but it seems like the reason why we were having some problems on adding some tests before this. This is not enabled by default, to use it I add a mutateEntry property on the azureRecording. I did this on the recordings for the subscriptions.


test/helpers/recording.ts
```diff
@@ -134,29 +149,6 @@ export function getMatchRequestsBy({
 }
 
 export function defaultShouldReplaceSubscriptionId(pathname: string): boolean {
```
ndowmon, line 151: I think it's fair to get rid of this function altogether, and always replace `subscriptionId` in the recording.

test/helpers/recording.ts
```diff
@@ -60,6 +60,21 @@ function mutateAccessToken(
   }
 }
 
+export function mutateSubscriptionAndDirectory(entry: RecordingEntry, config) {
+  mutateRecordingEntry(entry);
+  if (!entry.response.content.text || !entry.request.url) {
```
ndowmon, line 65: Let's make this two separate conditions, and do replacements for each.

test/helpers/recording.ts
```diff
@@ -60,6 +60,21 @@ function mutateAccessToken(
...
+    return;
+  }
+
+  entry.request.url = entry.request.url
+    .replace(new RegExp(`${config.directoryId}`, 'g'), 'directory-id')
+    .replace(new RegExp(`${config.subscriptionId}`, 'g'), 'subscription-id');
+
+  entry.response.content.text = entry.response.content.text
+    .replace(new RegExp(`${config.directoryId}`, 'g'), 'directory-id')
+    .replace(new RegExp(`${config.subscriptionId}`, 'g'), 'subscription-id');
```
ndowmon, line 75:
```suggestion
    .replace(new RegExp(config.directoryId, 'g'), 'directory-id')
    .replace(new RegExp(config.subscriptionId, 'g'), 'subscription-id');

  entry.response.content.text = entry.response.content.text
    .replace(new RegExp(config.directoryId, 'g'), 'directory-id')
    .replace(new RegExp(config.subscriptionId, 'g'), 'subscription-id');
```

src/steps/resource-manager/subscriptions/executionHandlers/subscriptionDiagnosticSettings.ts
```diff
@@ -0,0 +1,41 @@
...
+} from '../../../../types';
+import { steps as storageSteps } from '../../storage/constants';
+import {
+  createDiagnosticSettingsEntitiesAndRelationshipsForResource,
+  diagnosticSettingsEntitiesForResource,
+  getDiagnosticSettingsRelationshipsForResource,
+} from '../../utils/createDiagnosticSettingsEntitiesAndRelationshipsForResource';
+import { entities, steps } from '../constants';
+
+export async function fetchSubscriptionDiagnosticSettings(
```
ndowmon, line 13: Let's move this into `/subscriptions.ts`

src/steps/resource-manager/subscriptions/executionHandlers/subscriptions.test.ts
```diff
@@ -0,0 +1,37 @@
...
+  }
+});
+test('rm-subscription', async () => {
+  const stepTestConfig = getStepTestConfigForStep(steps.SUBSCRIPTION);
+
+  recording = setupAzureRecording({
+    name: steps.SUBSCRIPTION,
+    directory: __dirname,
+    mutateEntry: (entry) =>
+      mutateSubscriptionAndDirectory(entry, stepTestConfig.instanceConfig),
```
ndowmon, line 27: If possible, let's just bring this mutation into the standard` mutateEntry()`.
```suggestion
```

ndowmon (approved):

PR #562, opened 2023-03-22
Added some new error parsing and other checks on entity creation

Gonzalo-Avalos-Ribas:
Should reduce the amount of errors we are logging with undefined messages and in general those that are of the form `The current subscription type is not permitted to perform operations on any provider namespace. Please use a different subscription.`

### DIFF
--- a/src/azure/resource-manager/client.ts
+++ b/src/azure/resource-manager/client.ts
@@ -279,6 +279,13 @@ export async function request<T extends ResourceResponse>(
       if (err instanceof AzureRestError) {
         status = err.body?.code ?? err.response?.status;
         statusText = err.body?.message ?? err.body?.error?.message;
+        if (!statusText) {
+          try {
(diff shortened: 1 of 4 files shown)


src/azure/resource-manager/client.ts
```diff
@@ -279,6 +279,13 @@ export async function request<T extends ResourceResponse>(
       if (err instanceof AzureRestError) {
         status = err.body?.code ?? err.response?.status;
         statusText = err.body?.message ?? err.body?.error?.message;
+        if (!statusText) {
+          try {
+            statusText = JSON.parse(JSON.parse(err.message)).error.message;
```
ndowmon, line 284: This `JSON.parse(JSON.parse(err.message))` looks funky to me. Required?

Gonzalo-Avalos-Ribas: It is! Don't know why the message comes strigified twice, but i does

PR #566, opened 2023-04-13
Bump @azure/arm-storage@^15.3.0 to get allowSharedKeyAccess property

ndowmon:
Bumped `@azure/arm-storage` in order to add the `allowSharedKeyAccess` property to `azure_storage_account` entities.

---
Also modernized unit tests in the `storage` service to use the best-practice patterns. E.g.:

```ts
test('rm-storage-resources', async () => {
  const stepTestConfig = getStepTestConfigForStep(steps.STORAGE_ACCOUNTS);

  recording = setupAzureRecording(
    {
      name: steps.STORAGE_ACCOUNTS,
      directory: __dirname,
    },
    stepTestConfig.instanceConfig,
  );

  const stepResults = await executeStepWithDependencies(stepTestConfig);
  expect(stepResults).toMatchStepMetadata(stepTestConfig);
}, 20_000);
```


src/steps/resource-manager/storage/converters.ts
```diff
@@ -100,7 +100,7 @@ export function createStorageAccountEntity(
             ? encryptedServices.queue?.enabled
             : undefined,
         allowBlobPublicAccess: data.allowBlobPublicAccess,
-        allowSharedKeyAccess: (data as any).allowSharedKeyAccess,
+        allowSharedKeyAccess: data.allowSharedKeyAccess,
```
ndowmon, line 103: The new types include `allowSharedKeyAccess`, so no need to use `(data as any)` anymore.

src/steps/resource-manager/storage/index.test.ts
```diff
@@ -1,389 +1,90 @@
...
-
-    await fetchAccount(setupContext);
-    const accountEntities = setupContext.jobState.collectedEntities.filter(
-      (e) => e._type === ACCOUNT_ENTITY_TYPE,
-    );
-    expect(accountEntities.length).toBe(1);
-    const accountEntity = accountEntities[0];
-
-    await fetchKeyVaults(setupContext);
+test('rm-storage-resources', async () => {
```
ndowmon, line 17: Modernized all of the unit tests here. @Gonzalo-Avalos-Ribas had actually gone through many services in this project already to modernize, so I was able to leverage the tools & pattern he had already started using. 🎉

ndowmon: I removed the client & snapshot tests as this represents what we should really be asserting for integrations.

austinkelleher (approved):

PR #568, opened 2023-04-17
Consolidating Job Event Names to SDK supported values

VDubber:
- update job event names to SDK supported values
- format changes

This is the first part in providing users with a better indication of an integrations health.


.github/workflows/codeql-analysis.yml
VDubber, line 1: prettier

austinkelleher (approved):

PR #569, opened 2023-04-19
Ingest scale sets

Gonzalo-Avalos-Ribas:
Added ingestion of azure scale sets [Api docs](https://learn.microsoft.com/en-us/rest/api/compute/virtual-machine-scale-sets/list-all?tabs=HTTP).
Maybe we should relate this entity to others in compute?


src/steps/resource-manager/compute/constants.ts
```diff
@@ -39,6 +40,9 @@ export const VIRTUAL_MACHINE_IMAGE_ENTITY_CLASS = ['Image'];
 export const DISK_ENTITY_TYPE = 'azure_managed_disk';
 export const DISK_ENTITY_CLASS = ['DataStore', 'Disk'];
 
+export const VIRTUAL_MACHINE_SCALE_SET_ENTITY_TYPE = 'azure_vm_scale_set';
+export const VIRTUAL_MACHINE_SCALE_SET_ENTITY_CLASS = ['Deployment', 'Group'];
```
ndowmon, line 44: Nice, looks like this matches `aws_autosacling_group`: https://github.com/JupiterOne/graph-aws/blob/main/src/steps/services/autoscaling/constants.ts#L61-L64 👍

src/steps/resource-manager/compute/constants.ts
```diff
@@ -75,6 +79,11 @@ export const entities = {
     _type: VIRTUAL_MACHINE_ENTITY_TYPE,
     _class: VIRTUAL_MACHINE_ENTITY_CLASS,
   },
+  VIRTUAL_MACHINE_SCALE_SET: {
+    resourceName: '[RM] Virtual Machine Scale Set',
+    _type: VIRTUAL_MACHINE_SCALE_SET_ENTITY_TYPE,
+    _class: VIRTUAL_MACHINE_SCALE_SET_ENTITY_CLASS,
```
ndowmon, line 85: I recommend setting `_type` and `_class` here, to avoid temptation to import `VIRTUAL_MACHINE_SCALE_SET_ENTITY_TYPE` in other files.

You haven't done this, but this becomes easier if `_type` and `_class` have their own exported variables.

```diff
  return createIntegrationEntity({
    entityData: {
      source: data,
      assign: {
        _key: data.id!,
-       _type: VIRTUAL_MACHINE_SCALE_SET_ENTITY_TYPE,
+       _type: entities.VIRTUAL_MACHINE_SCALE_SET._type,
-       _class: VIRTUAL_MACHINE_SCALE_SET_ENTITY_CLASS,
+       _class: entities.VIRTUAL_MACHINE_SCALE_SET._class,
```

```suggestion
    _type:'azure_vm_scale_set',
    _class: ['Deployment', 'Group'],
```

src/steps/resource-manager/compute/converters.ts
```diff
@@ -276,7 +277,121 @@ export function createSharedImageVersion(
...
+  webLinker: AzureWebLinker,
+  data: VirtualMachineScaleSet,
+) {
+  return createIntegrationEntity({
+    entityData: {
+      source: data,
+      assign: {
+        _key: data.id!,
+        _type: entities.VIRTUAL_MACHINE_SCALE_SET._type,
+        _class: ['Deployment', 'Group'],
```
ndowmon, line 290:
```suggestion
        _class: entities.VIRTUAL_MACHINE_SCALE_SET._class,
```

src/steps/resource-manager/compute/converters.ts
```diff
@@ -276,7 +277,121 @@ export function createSharedImageVersion(
...
+        maxBatchInstancePercent:
+          data.upgradePolicy?.rollingUpgradePolicy?.maxBatchInstancePercent,
+        maxUnhealthyInstancePercent:
+          data.upgradePolicy?.rollingUpgradePolicy?.maxUnhealthyInstancePercent,
+        maxUnhealthyUpgradedInstancePercent:
+          data.upgradePolicy?.rollingUpgradePolicy
+            ?.maxUnhealthyUpgradedInstancePercent,
+        pauseTimeBetweenBatches:
+          data.upgradePolicy?.rollingUpgradePolicy?.pauseTimeBetweenBatches,
+        proximityPlacementGroupId: data.proximityPlacementGroup?.id,
```
ndowmon, line 319: Maybe in the future if we ingest proximity placement groups, we can build a relationship using this ID.

JuanManuelIbarlucea (approved):

src/steps/resource-manager/compute/converters.ts
```diff
@@ -276,7 +277,121 @@ export function createSharedImageVersion(
...
+        'virtualMachineProfile.scheduledEventsProfile.terminateNotificationProfile.enable':
+          data.virtualMachineProfile?.scheduledEventsProfile
+            ?.terminateNotificationProfile?.enable,
+        'virtualMachineProfile.scheduledEventsProfile.terminateNotificationProfile.notBeforeTimeout':
+          data.virtualMachineProfile?.scheduledEventsProfile
+            ?.terminateNotificationProfile?.notBeforeTimeout,
+        'virtualMachineProfile.storageProfile.dataDisks': data.virtualMachineProfile?.storageProfile?.dataDisks
+          ?.map((disk) => disk.name)
+          .filter((e) => !!e) as string[],
+        'virtualMachineProfile.storageProfile.imageReference.id':
```
ndowmon, line 376: Maybe in a follow-up PR, you can build a relationship to the image reference, if we have a resource to connect?

src/steps/resource-manager/compute/converters.ts
```diff
@@ -276,7 +277,121 @@ export function createSharedImageVersion(
...
+        'virtualMachineProfile.storageProfile.imageReference.exactVersion':
+          data.virtualMachineProfile?.storageProfile?.imageReference
+            ?.exactVersion,
+        'virtualMachineProfile.storageProfile.imageReference.sku':
+          data.virtualMachineProfile?.storageProfile?.imageReference?.sku,
+        'virtualMachineProfile.storageProfile.imageReference.version':
+          data.virtualMachineProfile?.storageProfile?.imageReference?.version,
+
+        webLink: webLinker.portalResourceUrl(data.id),
+      },
```
ndowmon, line 391: Which other relationships do you think we ought to build? 

I don't see any reference to `azure_vm`s here, but I suspect there might be a property on `azure_vm` that links back to this. That definitely seems like a valuable relationship to add!

ndowmon (approved):
Totally awesome. I've made a few suggestions on the `constants.ts` file, and a few suggestions of relationships that you should build in the future. All in all, though, looks great!

PR #570, opened 2023-04-24
Int 7422 disable steps legacy subscriptions

Gonzalo-Avalos-Ribas:
Disabled all the steps for subscriptions named 'Access to Azure Active Directory'
These subscriptions do NOT host Azure AD. These are legacy subscriptions that can no longer be managed by customer portal. We decided not to execute any step, just show an event.
 If (almost) any step executes we get:
` The current subscription type is not permitted to perform operations on any provider namespace. Please use a different subscription.`
  More info: [legacy subscriptions](https://www.jasonfritts.me/2020/04/07/what-is-the-access-to-azure-active-directory-subscription-for/#:%7E:text=The%20%E2%80%9CAccess%20to%20Azure%20Active%20Directory%E2%80%9D%20subscriptions%20are%20a%20legacy,portal.azure.com)

### DIFF
--- a/src/getStepStartStates.ts
+++ b/src/getStepStartStates.ts
@@ -2,6 +2,8 @@ import {
   IntegrationExecutionContext,
   StepStartStates,
   StepStartState,
+  IntegrationInfoEventName,
+  IntegrationWarnEventName,
 } from '@jupiterone/integration-sdk-core';
 
 import { IntegrationConfig } from './types';
@@ -113,6 +115,7 @@ import { PolicyInsightSteps } from './steps/resource-manager/policy-insights/con
 import { ManagementGroupSteps } from './steps/resource-manager/management-groups/constants';
 import { STEP_RM_CONTAINER_SERVICES_CLUSTERS } from './steps/resource-manager/container-services/constants';
 import { FrontDoorStepIds } from './steps/resource-manager/frontdoor/constants';
+import { J1SubscriptionClient } from './steps/resource-manager/subscriptions/client';
 
 function makeStepStartStates(
   stepIds: string[],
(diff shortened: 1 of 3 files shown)


src/getStepStartStates.ts
```diff
@@ -268,14 +271,50 @@ export function getResourceManagerSteps(): GetApiSteps {
...
+      );
+      if (
+        subscription &&
+        subscription.displayName == 'Access to Azure Active Directory'
+      ) {
+        disableAllSteps = true;
+        logger.publishInfoEvent({
+          name: IntegrationInfoEventName.Results,
+          description:
+            'The subscription is is not permitted to perform operations on any provider namespace. These are legacy subscriptions that can no longer be managed by customer portal. No steps will be executed.',
```
ndowmon, line 299: I would suggest something like this:
```suggestion
          description:
            'This is an "Access to Azure Active Directory" subscription, which is a legacy subscription type that is no longer used. They were used prior to the current Azure Portal (https://portal.azure.com/). JupiterOne cannot ingest any resources from this subscription.',
```

ndowmon: I'm also wondering if we should update our `configureSubscriptionInstances()` logic to stop ingesting these subscriptions to not even pull these in to begin with. 

```diff
function shouldSkipSubscription(subscription: Subscription) {
  return subscription.state !== 'Enabled' 
  || subscription.tags?.JupiterOne === 'SKIP'
+ || subscription.displayName === 'Access to Azure Active Directory'
}
```

ndowmon (dismissed):

src/getStepStartStates.ts
```diff
@@ -268,14 +271,50 @@ export function getResourceManagerSteps(): GetApiSteps {
...
+  if (hasSubscriptionId(config)) {
+    const subscriptionClient = new J1SubscriptionClient(config, logger);
+    try {
+      const subscription = await subscriptionClient.getSubscription(
+        config.subscriptionId!,
+      );
+      if (
+        subscription &&
+        subscription.displayName == 'Access to Azure Active Directory'
+      ) {
```
ndowmon, line 294: Curious - what if `subscription` or `subscription.displayName` is undefined? It seems like `getSubscription()` returns `Subscription | undefined`.

Maybe make us aware of that case:
```suggestion
      if (!subscription?.displayName) {
        logger.warn({ subscriptionId: config.subscriptionId }, 'Could not get subscription for subscriptionId');
      } else if (subscription.displayName === 'Access to Azure Active Directory') {
```

ndowmon (approved):

PR #601, opened 2023-10-10
Add factor to retries - Max 20 retries for auth scope

Gonzalo-Avalos-Ribas:


docs/jupiterone.md
```diff
@@ -414,21 +414,24 @@ The following relationships are created:
```
Gonzalo-Avalos-Ribas, line 414: No changes here - just re ordering

RonaldEAM (approved):

PR #617, opened 2024-01-18
Int 10063 remove converter and client tests

Gonzalo-Avalos-Ribas:
Remove All converter tests, and most of client tests. Left only the ones that try failed states such as 400s or 500s.


src/steps/resource-manager/.DS_Store
RonaldEAM, line 1: Is this needed?

RonaldEAM (approved):

PR #635, opened 2024-04-12
completed rolebinding step and relationship

poornima-metron:


src/steps/resource-manager/container-services/converters.ts
```diff
@@ -23,6 +23,16 @@ export function getAccessRoleKey(name, location) {
   return `azure_access_role_${name}_location_${location} `;
 }
 
+function getEntityFromId(id: string, entityName): string {
+  const parts = id.split('/');
+  const index = parts.indexOf(entityName);
+  if (index !== -1 && index + 1 < parts.length) {
+    return parts[index + 1];
+  } else {
+    throw new Error('Invalid id format');
```
suraj-metron, line 32: Dont throw an error. Wrap the function in try catch. If there is any exception, just log it and return a default value or empty string maybe

src/steps/resource-manager/container-services/index.ts
```diff
@@ -82,6 +84,54 @@ export async function fetchMaintenanceConfigurations(
...
+        async (rolebinding) => {
+          const rolebindingEntity = createRoleBindingEntity(
+            webLinker,
+            rolebinding,
+          );
+          if (!jobState.hasKey(rolebindingEntity._key)) {
+            await jobState.addEntity(rolebindingEntity);
+          }    
+          const relationship = createMappedRelationship({
+            _key: `${rolebindingEntity._key}|IS|ClusterRole:${rolebindingEntity._key}`,
```
suraj-metron, line 109: Use a function to return this value

suraj-metron (changes requested):

src/steps/resource-manager/container-services/converters.ts
poornima-metron: Wrapped the function in try catch

src/steps/resource-manager/container-services/index.ts
poornima-metron: created function to generate key

PR #637, opened 2024-04-29
adding props to entity

mikiodehartj1:
adding the following props:
nodeImageVersion
osType
osSku


src/steps/resource-manager/container-services/converters.ts
```diff
@@ -34,6 +34,9 @@ export function createClusterEntitiy(
         enablePodSecurityPolicy: data.enablePodSecurityPolicy,
         disableLocalAccounts: data.disableLocalAccounts,
         webLink: webLinker.portalResourceUrl(data.id),
+        nodeImageVersion: data.agentPoolProfiles?.map(i => i.nodeImageVersion as string),
```
VDubber, line 37: What is the reason for casting `nodeImageVersion` to string?

i5o: I helped Mikio with this, but the reason is the same as we do in aws, agentPoolProfiles can `x[] | undefined`, so for type safety TS forces us to use `as string`

image

VDubber (approved):

PR #646, opened 2024-05-24
Azure access group dev

poornima-metron:


src/steps/resource-manager/access-groups/client.ts
```diff
@@ -0,0 +1,231 @@
+import { IntegrationWarnEventName } from '@jupiterone/integration-sdk-core';
+import {
```
Gonzalo-Avalos-Ribas, line 2: All steps that use the graph api or that dont depend on a subscription should go inside the folder of active-directory instead of resource-manager. Can we please create a folder inside of active-direcory that is called access-groups and move the code there?

suraj-metron: Done

Gonzalo-Avalos-Ribas (approved):

PR #659, opened 2024-08-05
APP-15543 - Use createIntegrationHelpers

jzolo22:


src/steps/active-directory/conditional-access-policy/entities.ts
```diff
@@ -0,0 +1,110 @@
...
+    ),
+    countryLookupMethod: SchemaType.Optional(SchemaType.String()),
+    public: SchemaType.Boolean({
+      description:
+        'This value is not applicable to this entity, however is expected on a entity with a _class of `Network`',
+      default: false,
+    }),
+    internal: SchemaType.Boolean({
+      description:
+        'This value is not applicable to this entity, however is expected on a entity with a _class of `Network`',
```
jzolo22, line 92: Not necessarily thrilled with this but couldn't think of a better way to handle

src/steps/active-directory/conditional-access-policy/converter.ts
```diff
@@ -105,19 +110,20 @@ export function createConditionalAccessNamedLocationEntity(data: any): Entity {
...
+        createdDateTime: parseTimePropertyValue(data.createdDateTime),
+        createdOn: parseTimePropertyValue(data.createdDateTime),
         countriesAndRegions: data.countriesAndRegions,
         countryLookupMethod: data.countryLookupMethod,
-      },
+        // TODO: Determine if `Network` is the correct _class for this entity given the schema & fix if necessary
+        public: false,
+        internal: false,
+        CIDR: null,
+      }),
```
jzolo22, line 126: Same with comment below - not sure this is the best way to handle these required properties...

PR #661, opened 2024-08-06
Kno 427

SeaBlooms:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@jupiterone/graph-azure",
-  "version": "7.6.0",
```
Gonzalo-Avalos-Ribas, line 3: There is no need to modfiy this by hand just add the tags of minor to the pr

SeaBlooms: ok gotcha, reverted to 7.6.0

Gonzalo-Avalos-Ribas (approved):

PR #662, opened 2024-08-06
add more record type properties

SeaBlooms:


src/steps/resource-manager/dns/converters.ts
```diff
@@ -52,9 +52,28 @@ export function createDnsRecordSetEntity(
...
+        // SOA type
+        soaRecordHost: data.soaRecord?.host,
+        soaRecordEmail: data.soaRecord?.email,
+        soaRecordserialNumber: data.soaRecord?.serialNumber,
+        // A type
+        aRecords: data.aRecords,
+        // AAAA type
+        aaaaRecords: data.aaaaRecords,
+        // NS type
+        nsRecords: data.nsRecords,
```
VDubber, line 68: what is the data type of these values? array of string?

VDubber: objects / dictionaries are not supported!

if nsRecords is an object like:
```
{ val: "one", val2: "two" }

```

this will not be accepted by the data pipeline.

Gonzalo-Avalos-Ribas (approved):

PR #671, opened 2024-08-28
Add directory id to subscription

Gonzalo-Avalos-Ribas:
Added tenantId property in azure_subscription Entity
Jira Ticket: https://jupiterone.atlassian.net/browse/IA-172

### DIFF
--- a/src/steps/active-directory/converters.ts
+++ b/src/steps/active-directory/converters.ts
@@ -139,6 +139,7 @@ export function createAccountEntityWithOrganization(
         defaultDomain,
         verifiedDomains,
         securityDefaultsEnabled: securityDefaults?.isEnabled,
+        vendor: organization.displayName,
       },
     },
   });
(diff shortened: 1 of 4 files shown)


src/steps/active-directory/converters.ts
```diff
@@ -139,6 +139,7 @@ export function createAccountEntityWithOrganization(
         defaultDomain,
         verifiedDomains,
         securityDefaultsEnabled: securityDefaults?.isEnabled,
+        vendor: organization.displayName,
```
VDubber, line 142: This should always equal "Microsoft". The vendor of the provider API.

src/steps/resource-manager/subscriptions/converters.ts
```diff
@@ -34,6 +35,8 @@ export function createSubscriptionEntity(
         ...tags,
         webLink: webLinker.portalResourceUrl(data.id),
         offerName: data.subscriptionPolicies?.quotaId,
+        tenantId: directoryId,
+        vendor: data.displayName,
```
VDubber, line 39: Is a subscription really a class type of Account??
As with my other vendor comment, this should be Microsoft. See [here](https://github.com/JupiterOne/jupiter-data-model/blob/main/packages/jupiterone-data-model/src/class_schemas/Account.json#L22)

VDubber (approved):

PR #672, opened 2024-09-02
INT-11596: add public access property for database servers

gastonyelmini:


src/steps/resource-manager/databases/sql/converters.ts
```diff
@@ -147,8 +147,7 @@ export function createSqlServerFirewallRuleEntity(firewallRule: FirewallRule) {
         name: firewallRule.name,
         id: firewallRule.id,
         category: ['host'],
-        kind: firewallRule.kind,
-        location: firewallRule.location,
+        type: firewallRule.type,
```
gastonyelmini, line 150: Unfortunately, this properties were removed in the new version of the SDK,  but the `type` one was added so i'm using it here.

Gonzalo-Avalos-Ribas: Maybe we should add some release notes to make sure that users know about this.

Gonzalo-Avalos-Ribas (approved):

